PR #14, opened 2016-11-03
Milan 2016 Codemotion preparation

mauroservienti:
What's in there:

* changes S.P.A. HTTP port to prevent browser caching issues
* upgrades to latest NSB
* fixes page title on all samples
* Merge PreReq checker source from master
* aligns NET FX version to 4.6.1
* removes wrong project ref to IT/Ops in Pub/Sub
* NServiceBus not needed in 1st exercise
* renames solutions to simplify having more than one open at a time
* Updates licenses to a new trial that ends on January, 15 2017


adamralph:
@mauroservienti you've included .suo files in the commit was that intentional?

Also I think the `.vs` folders shouldn't be committed? They contain environment specific data, such as the absolute file path to your clone. AFAIK these are auto-generated when you build/run the projects in VS.

mauroservienti:
yes, intentional so that we have control over the http port settings for web projects, it has been like that from the very beginning IIRC

adamralph:
OK, I guess that's an argument for including the `.vs` folders, or at least the content within them that controls that setting.

What about the .suo files? It looks like they've been added in these commits.

mauroservienti:
IIRC they are required to guarantee that startup projects are persisted.

mauroservienti:
Re the startup projects nightmare this extension seems to be very nice: https://visualstudiogallery.msdn.microsoft.com/f4e1be8c-b2dd-4dec-b273-dd88f8818571

Should we move to that instead of committing the `.suo` files?

adamralph:
The VS gallery extension is a personal tooling choice. If I were in the workshop I'd probably choose not install it and just use the regular UI (even though it's poor).

For a workshop repo, perhaps it's OK to commit the `.suo` files. After all, it's directed learning rather than something for people to "work on".

mauroservienti:
@Particular/microservices-workshop-maintainers please review and merge

dvdstelt:
@mauroservienti 22 commits and 384 files changed, without a lot of clarification in the topic description on why there were changes and what they are? I can understand some of it via the comments, but it's hardly reviewable.

But I'll have a look and see if everything works on my machine, so we can put a sticker on it and merge! :-)

mauroservienti:
Commits messages seem pretty clear to me. Mostly dependencies, no code changes except what is required by latest NSB. Some tweaks to fix caching issues we had in Oslo. Feel free to ask.

PR #26, opened 2016-12-28
[WIP] Replaces client side events with visitor pattern

mauroservienti:
this is intended to suggest to replace the actual approach used to load composite data from different services moving from a client side pub/sub mechanism to a visitor pattern.

In order to retrieve Orders with Customers details the current flow is:

1. At startup
   1. Sales registers itself as the module responsible to retrieve `orders` data
   1. All others modules subscribes to events
1. Query Sales service to get the Orders list
   1. Each order in the above list has a customer id property
1. Once the Orders list is retrieved a client side event is published
1. Customer module is subscribed to the event
   1. Aggregates customer ids
   1. Queries the customers service
1. Once the response is back data are composed in the UI

The problem with the above approach is that the even is broadcasted at step `3` and subscribers kicks in asynchronously, this makes it impossible for the UI to wait for all the modules to have retrieved their data before composing the UI.

The proposed approach solves the above problem in the following way. In order to retrieve Orders with Customers details the proposed flow is:

1. At startup
   1. Sales registers itself as the module responsible to retrieve `orders` data
   1. _All others modules interested in the orders queries register themselves as query results visitors_
1. Query Sales service to get the Orders list
   1. Each order in the above list has a customer id property
1. _Once the Orders list is retrieved all visitors are given the opportunity to chime in and visit the results_
1. Customer module is _registered as visitor_
   1. Aggregates customer ids
   1. Queries the customers service
1. Once the response is back data are composed in the UI

(In _italic_ the differences)

Now step 3 can be async or sync as we prefer allowing a better UI composition flow and a simpler approach.

@dvdstelt @ramonsmits @Particular/microservices-workshop-maintainers let's have a quick chant about this in Lisbon while we discuss the whole client side implementation.

**Once approved and merged it needs to be ported to all the exercises.**


mauroservienti:
Once https://github.com/Particular/Workshop.Microservices/pull/30 is in place this has to be revisited.

PR #27, opened 2016-12-28
[WIP] Removes templating engine to simplify the mechanics of the UI

mauroservienti:
The actual mechanics of the UI Composition are based on a `templates engine` that is not really required. It only complicates things for attendees that other then getting all the AngularJS details need also to dive into a very complex AngularJS directive.
This PR entirely removes the need for the `templates engine` simplifying the mechanics of the UI.

@dvdstelt @ramonsmits @Particular/microservices-workshop-maintainers let's have a quick chant about this in Lisbon while we discuss the whole client side implementation.

**Once approved and merged it needs to be ported to all the exercises.**


mauroservienti:
Once https://github.com/Particular/Workshop.Microservices/pull/30 is in place this has to be revisited.

PR #30, opened 2017-01-18
Mvc basic demo

mauroservienti:
first demo as discussed with @udidahan and @dvdstelt implementing ViewModel composition in ASP.Net MVC using an `IResultFilter` to inject the composition logic into the MVC request Pipeline.

To keep things as simple as possible for attendees the MVC front-end has no DI at all and [filters are hardcoded](https://github.com/Particular/Workshop.Microservices/blob/5531d266f3226d4895ffd49e88574be51a5327c9/demos/CompositeUI-MVC/Divergent.Frontend/App_Start/FilterConfig.cs#L13) do we want to have full DI support in the front-end as well?


dvdstelt:
I would not bother with stuff like DI and so on. We need to get this (and other stuff) ready before Wednesday! 😅

mauroservienti:
agree, so let's review this in today's meeting and merge if we're happy

mauroservienti:
just reviewed this with @dvdstelt and @udidahan on a call. Merging.

PR #35, opened 2017-01-28
ProcessOrder method must be called from Handle methods

ramonsmits:
Missing part in exercise 03. ProcessOrder method must be called from Handle methods.


mauroservienti:
@ramonsmits I merged it into `exercise-2-new-design` that is where I'll branch off for to apply changes for exercise 3

PR #38, opened 2017-01-31
Async db call in the pub-sub example

serbrech:


mauroservienti:
@serbrech thanks. Have you actually checked it works as expected? When I first tested it the first time, some months ago, the SQLite drive wasn't very happy.

serbrech:
ahaa, I'll verify.
just looked suspicious that Save was async, but not the get.

PR #40, opened 2017-01-31
Type conflict with NServiceBus.Order

tjeerdhans:


mauroservienti:
Much better thanks, otherwise one needs to deal with using statements and weird compilation errors.

PR #42, opened 2017-02-01
Update 03 - Sagas.md

serbrech:
The description of the `ProcessOrder` methods made it look like it was a hook from NServiceBus Saga being called by the framework.
I added a precision to tell that it has to be called from both `Handle` methods.


mauroservienti:
thanks

PR #43, opened 2017-02-01
[WIP] Spike: Ensure SQLite migrations are run only by one project

mauroservienti:
@dvdstelt this is a very raw way to prevent concurrency in creating SQLite dbs at startup time. Thoughts?
[Relevant bits](https://github.com/Particular/Workshop.Microservices/blob/26ce6c942bd79b296224c4603f6cc59dcf6e4526/exercises/src/01%20CompositeUI/after/Divergent.Sales.Data/Context/SalesContext.cs#L25)

Fix https://github.com/Particular/Workshop.Microservices.Private/issues/38


mauroservienti:
dropping this in favor of moving all exercises to (localdb)

PR #44, opened 2017-02-01
[WIP] Spike: Ensures concurrency is handled when dealing with SQLite

mauroservienti:
@dvdstelt in this configuration all the failure scenarios we faced in Amsterdam are handled. Honestly it is so horrible that I'd be more than happy to evaluate another storage solution for samples/exercises.


serbrech:
Is SqlServer Compact Edition better?

dvdstelt:
@serbrech We are definitely looking into something different. Ideally people should not have to install anything. I'm still not sure if LocalDb is installed with Visual Studio 2015 or not...

mauroservienti:
dropping this in favor of moving all exercises to (localdb)

PR #45, opened 2017-02-01
Update 04 - Integration with ITOps.md

mhoppeats:
CreateFedexRequest returned a null value instead of an XDocument.
Not a big problem in the exercise as the return value is not used.


mauroservienti:
Good catch. Thanks.

PR #53, opened 2017-03-27
Remove Task.Run usage in view model appender

danielmarbach:
@mauroservienti ping


mauroservienti:
I spoke too early, it's failing on my machine. I end up with a null view model in the Razor template.

PR #59, opened 2017-04-15
ASP.NET Core sample

mauroservienti:
## PoA

- [x] define custom routes
- [x] allow users to define their routes
- [x] inject custom routs into the default routing engine
- [x] gateway pipeline terminator
- [x] support GET requests
- ~support POST requests~
- ~support PUT requests~
- ~support PATCH requests~
- ~support DELETE requests~
- [x] add sales and shipping samples (with hardcoded data)
- [x] WebAPI hosts to expose real sales and shipping data
- [x] Mimic the same pure MVC sample as in the first Demo
- ~super simple SPA to interact with back-end~
- [x] is this an API Gateway or a reverse proxy? Once defined rename accordingly.
- [x] review with Udi
- [x] split into 2 solutions each one loading one single frontend
- ~In the solution that demos the Gateway alone add a super-simple js client~ replaced by https://github.com/Particular/Workshop.Microservices/pull/62

## Description

This PR introduces a demo of a composite UI using as composition infrastructure a gateway/reverse proxy built on top of ASP.Net Core 1.1.

This composition infrastructure allows to run a .Net Core web server as simple as:

```
public class Startup
{
    public void ConfigureServices(IServiceCollection services)
    {
        services.AddRouting();
        services.AddViewModelComposition();
    }

    public void Configure(IApplicationBuilder app, ILoggerFactory loggerFactory)
    {
        app.RunViewModelCompositionWithDefaultRoutes();
    }
}
```

The above configures a default route, whose template is `{controller}/{id:int?}`, that mimics the WebAPI default behavior. All HTTP requests are intercepted and if matching `ViewModelAppenders` are found they are invoked, allowing composition to happen.
Similarly to the full MVC Classic sample lists are handled via subscribing to events raised by appenders.


mauroservienti:
This is ready for review, I know it's huge but I have no other idea on how to commit such a large sample other than this type of PRs (suggestions are more than welcome).

BTW take a look at the PoA, there is a pending question:

> is this an API Gateway or a reverse proxy? Once defined rename accordingly.

demos/ASP.Net Core API Gateway/ITOps.ViewModelComposition/CompositionHandler.cs
```diff
@@ -0,0 +1,59 @@
...
+                    );
+                }
+
+                if (pending.Count == 0)
+                {
+                    return (null, StatusCodes.Status404NotFound);
+                }
+                else
+                {
+                    await Task.WhenAll(pending);
```
tmasternak: What happens when one of the appenders fails? Should we return partial response or fail the whole request?

demos/ASP.Net Core API Gateway/Divergent.Sales.ViewModelComposition/OrdersListViewModelAppender.cs
```diff
@@ -0,0 +1,63 @@
...
+             * this way to satisfy both the Gateway and the Mvc sample
+             */
+            var controller = (string)routeData.Values["controller"];
+
+            return HttpMethods.IsGet(verb)
+                && controller.ToLowerInvariant() == "orders"
+                && !routeData.Values.ContainsKey("id");
+        }
+
+        public async Task Append(dynamic vm, RouteData routeData, IQueryCollection query)
```
tmasternak: Should we pass vm here or change the approach so that the appender returns dynamic? This would prevent any cross-appender type of communication and name clashes between appenders.

In addition the structure of the end response could be controlled by the caller.

demos/ASP.Net Core API Gateway/Divergent.Sales.ViewModelComposition/OrdersListViewModelAppender.cs
```diff
@@ -0,0 +1,63 @@
...
+using System.Collections.Generic;
+using System.Dynamic;
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace Divergent.Sales.ViewModelComposition
+{
+    public class OrdersListViewModelAppender : IViewModelAppender
+    {
+        public bool Matches(RouteData routeData, string verb)
```
tmasternak: Could we provide some base class to make this more declarative?

tmasternak:
I like the approach! There are some high level questions that I had during the review which I would also like to ask:
 1. Did we consider the composition to be made on the basis of explicit template? Currently it's the `DynamicViewModel` which is a structure to which appenders and subscribers glue their fragments?
 2. What is the failure handling approach? Do we support partial reponses? If yes should we notify clients somehow?
 3. What is the assumed security model? Authorization by the Gateway and no verification later on?
 4. Is the url the only source of data for request handling? Should we enable/show how to get data from cookies as well?
 5. Does the caching leave only in the `CompositionHandler`?

demos/ASP.Net MVC/Divergent.Frontend/Views/Orders/Details.cshtml
```diff
@@ -16,12 +15,3 @@
...
 </div>
-
-@*<div class="panel panel-default">
-    <div class="panel-heading">
-        <h3 class="panel-title">Shipment information</h3>
-    </div>
-    <div class="panel-body">
-
-    </div>
-</div>*@
```
adamralph: @mauroservienti this file belongs to the existing solution, did this change creep into this PR by accident?

(Same with the .suo file above.)

demos/ASP.Net Core API Gateway/ITOps.ViewModelComposition/CompositionHandler.cs
mauroservienti: the whole thing blows up, we talked about it when the very first demo was built and decided that to keep the demo as simple as possible failure handling was not in scope.

demos/ASP.Net Core API Gateway/Divergent.Sales.ViewModelComposition/OrdersListViewModelAppender.cs
mauroservienti: yes and no, probably in a production ready thing would be a nice to have, in a sample not necessarily.

demos/ASP.Net Core API Gateway/Divergent.Sales.ViewModelComposition/OrdersListViewModelAppender.cs
mauroservienti: That's a good point. If appenders create their own VM that is returned the issue is that who manages the appender then needs to act as a coordinator in order to merge together all the returned view models. The result is that the coordinator knows the structure and we don't want that.

The used approach on the contrary makes the coordinator totally unaware of the structure, leaving the thing with the possibility that 2 appenders clash on the same property name for example.

demos/ASP.Net MVC/Divergent.Frontend/Views/Orders/Details.cshtml
mauroservienti: ooops. I made those changes while demoing the thing to you and @DavidBoike and then pushed them by accident.

mauroservienti: `.suo` files are there for the only purpose of sharing startup projects settings, so that we can provide a `F5` experience. I hate to be forced to commit that stuff.

Any other idea?

mauroservienti:
> 1. Did we consider the composition to be made on the basis of explicit template? Currently it's the `DynamicViewModel` which is a structure to which appenders and subscribers glue their fragments?

no, even if it's a thing that would be interesting to explorer. When I started this the goal was to demo a gateway-only composition to return json structures. But on top of that now I don't see why such an evolution should not be tested. @tmasternak can you expand a little bit on the idea? (maybe in its own issue)
 
> 2. What is the failure handling approach? Do we support partial reponses? If yes should we notify clients somehow?

There is no failure handling approach at all ATM.

>  3. What is the assumed security model? Authorization by the Gateway and no verification later on?

Appenders can retrieve security related headers from the incoming request and pass along to their own backend. I haven't thought about it all

>  4. Is the url the only source of data for request handling? Should we enable/show how to get data from cookies as well?

uhm. Interesting, tell me more

>  5. Does the caching leave only in the `CompositionHandler`?

in a large system one may end up with thousands of appenders, that's the first thing that needs to be cached since they can be easily cached by URL

mauroservienti:
@adamralph @seanfarmar since you saw the presentation at DDD SouthWest, thoughts on this:

>  is this an API Gateway or a reverse proxy? 

I'm on the fence, from the implementation perspective is a reverse proxy, conceptually is an API Gateway. The decisions affects the components names.

/cc @tmasternak @DavidBoike

tmasternak:
> no, even if it's a thing that would be interesting to explorer. When I started this the goal was to demo a gateway-only composition to return json structures. But on top of that now I don't see why such an evolution should not be tested. @tmasternak can you expand a little bit on the idea? (maybe in its own issue)

Will do. 

>uhm. Interesting, tell me more

I think you already answered my question when talking about security. Appenders already have access to the request so they can grab pretty much anything that client sends.

demos/ASP.Net Core API Gateway/Divergent.Sales.ViewModelComposition/OrdersListViewModelAppender.cs
tmasternak: First of all, I don't see that as a big issue but what the coordinator could do is:
a. simply append vm returned from appenders to the response (no knowledge about appenders needed)
b. detect clashes and do some namespacing (add a json-node derived from appender name etc.)

dvdstelt:
> is this an API Gateway or a reverse proxy

Isn't an API gateway an implementation of a reverse proxy? They're not mutually exclusive.
Since API Gateway is better explaining the implementation and better known, I'd go for API Gateway.

mauroservienti:
I have another question @dvdstelt @adamralph @tmasternak:

> Actually the demo supports GET requests only, in light of this request handlers were named `IViewModelAppender`. The semantic of POST, PUT, PATCH and DELETE doesn't align at all with the concept of "appending" data.
> It is much more request handling. I'm actually in favor of changing `IViewModelAppender` to `IHandleRequests` because from the HTTP transport perspective a GET and a POST are both requests.

Thoughts?

tmasternak:
@mauroservienti 👍 for using `IHandleRequets`.

Which leads us to other data source which are `ISubscribeToCompositionEvents`. 
First of all I was thinking if something along the lines of `IEnrichResponse` would be more appropriate Secondly I would really like to hear your opinions about coupling that we currently have between: http request, appenders and subscribers. Especially the subscribers are things that I would like to question. Here are the problems that I see:
 * They need to match request to work properly - I'm not sure if that is by design or working around framework behavior
 * It introduces strong coupling between artifacts that are owned by different teams (`Sales` owns `OrdersListViewModelAppender` and `Shipping` owns `OrdersLoadedSubscriber`). 
Are those indicators that the `template` approach with dedicated team creating template of responses and defining response composition is something we might want to PoC?

dvdstelt:
@mauroservienti Don't you think `IHandleRequests` sounds too generic and `IViewModelAppender` describes better what the purpose of a class does, implementing this interface? Or does it do other things?

I'll let @mauroservienti respond to @tmasternak his questions, because he knows best. But afaik most of it is because of the framework and the fact that it's really hard to come up with a good solution that doesn't involve writing TONS of interfaces and classes, without providing anything additional.

mauroservienti:
@tmasternak 

> Which leads us to other data source which are ISubscribeToCompositionEvents.

I haven't thought to `ISubscribeToCompositionEvents` yet, but your comment sounds reasonable 😉 

> and defining response composition is something we might want to PoC?

What about, starting from this, add a new "demo" or create a PoC?

mauroservienti:
> Don't you think IHandleRequests sounds too generic and IViewModelAppender describes better what the purpose of a class does, implementing this interface? Or does it do other things?

@dvdstelt I agree but it's not aligned with the HTTP behavior, so it might create a lot of confusion when it comes to strage HTTP usages such as POST requests to overcome GET URL length limitations. That means that you might want to have a `IViewModelAppender` invoked by a POST. As it stands now you cannot, appenders are bound to GET requests.
And HTTP has always an incoming URL + payload and always expects a response payload. Verbs are on top of the transport.

mauroservienti:
I'm playing with the new syntax [here](https://github.com/mauroservienti/shopper/tree/it-ops-composition/src/ITOps.ViewModelComposition) with the goal of building a slightly more real life sample that handles reads and writes.

mauroservienti:
Found a bug https://github.com/Particular/Workshop.Microservices/issues/64, I'll fix before merging.

seanfarmar:
@mauroservienti any chance you can do a screencast to give an overview of the solution? (similar to your presentation in DDD)?

mauroservienti:
Good idea @seanfarmar, I can do that this Friday

demos/ASP.Net Core API Gateway/Divergent.Sales.API.Host/Controllers/OrdersController.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Linq;
+using System.Web.Http;
+
+namespace Divergent.Sales.API.Host.Controllers
+{
+    [RoutePrefix("api/orders")]
+    public class OrdersController : ApiController
+    {
+        [HttpGet]
+        public dynamic Get(int id)
```
SimonCropp: why dynamic?

demos/ASP.Net Core API Gateway/ITOps.ViewModelComposition.Gateway/ComposableRouteHandler.cs
```diff
@@ -0,0 +1,46 @@
...
+    class ComposableRouteHandler
+    {
+        public static async Task HandleGetRequest(HttpContext context)
+        {
+            var result = await CompositionHandler.HandleGetRequest(context);
+            if (result.StatusCode == StatusCodes.Status200OK)
+            {
+                string json = JsonConvert.SerializeObject(result.ViewModel, GetSettings(context));
+                context.Response.ContentType = "application/json; charset=utf-8";
+                await context.Response.WriteAsync(json);
```
SimonCropp: this should be a direct return? ie no await

demos/ASP.Net Core API Gateway/ITOps.ViewModelComposition/Subscription.cs
```diff
@@ -0,0 +1,11 @@
+﻿using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
+using System.Threading.Tasks;
+
+namespace ITOps.ViewModelComposition
+{
+    abstract class Subscription
```
SimonCropp: why not an interface?

demos/ASP.Net Core API Gateway/Divergent.Sales.API.Host/Controllers/OrdersController.cs
mauroservienti: Good point, probably back in time the return type was an anonymous class an di we never changed the action return type, that I should always serilized.
I'll fix it. Thanks.

demos/ASP.Net Core API Gateway/ITOps.ViewModelComposition.Gateway/ComposableRouteHandler.cs
mauroservienti: Yeah, that `await` is useless

mauroservienti:
(rebased on master)

demos/ASP.Net Core API Gateway/Divergent.Sales.API.Host/Controllers/OrdersController.cs
mauroservienti: I checked and we return an anonymous type:

```
var order = db.Orders
    .Include(o => o.Items)
    .Where(o => o.Id == id)
    .Select(o => new
    {
        Number = o.Id,
        o.Id,
        ItemsCount = o.Items.Count
    })
    .SingleOrDefault();

return order;
```

in which case the return type can only be `dynamic` or `object`

demos/ASP.Net Core API Gateway/ITOps.ViewModelComposition.Gateway/ComposableRouteHandler.cs
mauroservienti: it can be, but it doesn't compile because at line 13 there is an `await`:

```
var result = await CompositionHandler.HandleGetRequest(context);
```

demos/ASP.Net Core API Gateway/ITOps.ViewModelComposition/Subscription.cs
mauroservienti: yeah, changed to be an interface.

demos/ASP.Net MVC/Divergent.Frontend/Views/Orders/Details.cshtml
adamralph: I think committing the .suo files is OK for a demo project, but the .sou for the ASP.NET MVC demo should not be changed as part of this PR. I'd prefer to have that change reverted in this branch (as well the change to this file).

mauroservienti: Having multiple demo solution should in the same repo has this side effect, not fun.
Which change/commit is "that" change?

adamralph: @mauroservienti I would like to revert the changes to these files:

- demos/ASP.Net MVC/.vs/ASP.Net MVC/v14/.suo
- demos/ASP.Net MVC/Divergent.Frontend/Views/Orders/Details.cshtml

I don't think a specific commit needs to be reverted. They can both be reverted to what's in `master`.

mauroservienti: You mean by manually adding back what I removed previously, correct?

adamralph: There are many ways to do it. Personally I'd open the TortoiseGit log view, then diff the tip of the branch to master, and then right click each file and "Revert to 2cbcea1".

mauroservienti:
fixed history, reverted a couple of wrong commits. Thanks to @adamralph for his support and git knowledge.

mauroservienti:
@adamralph, as per our call:

* demo is now split into 2 different solutions
   * `ASP.Net Core API Gateway`
   * `ASP.Net MVC Core`
* renamed the demo root folder from `ASP.Net Core API Gateway` to `ASP.Net Core`

In the coming days I'll have a look at the last item:

>  In the solution that demos the Gateway alone add a super-simple js client

Or is it better to merge this as is and raise a new issue for that, this is already big enough on its own

adamralph:
👍 thanks @mauroservienti 

Regarding the JS client for the gateway, I think it _can_ be done separately, since the gateway is a completely new aspect to the demo. However, I think you can start work on it now. If it gets done in time to go in with this PR then that's good, but if not, it can be sent in another PR.Perhaps you should push any WIP to a new branch based on this one. If this branch is not merged by the time the work is done. You can fast forward this branch to the tip of that new branch and include the JS client in this PR.

demos/ASP.Net Core/ITOps.ViewModelComposition.Gateway/ApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+            routes?.Invoke(routeBuilder);
+
+            app.UseRouter(routeBuilder.Build());
+        }
+
+        public static void RunCompositionGatewayWithDefaultRoutes(this IApplicationBuilder app)
+        {
+            app.RunCompositionGateway(routes =>
+            {
+                routes.MapComposableGet( template: "{controller}/{id:int?}");
```
adamralph: Extra space before `template`.

demos/ASP.Net Core/ITOps.ViewModelComposition.Gateway/ApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+                routes.MapRoute("{*NotFound}", context =>
+                 {
+                     context.Response.StatusCode = StatusCodes.Status404NotFound;
+                     return Task.CompletedTask;
+                 });
+            });
+        }
```
adamralph: This method can be expression-bodied.

demos/ASP.Net Core/ASP.Net Core API Gateway.sln
```diff
@@ -0,0 +1,107 @@
...
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Divergent.Sales.API.Host", "Divergent.Sales.API.Host\Divergent.Sales.API.Host.csproj", "{DC8ADA37-837C-490F-BDFC-FC8D1001417E}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Divergent.Shipping.Data", "Divergent.Shipping.Data\Divergent.Shipping.Data.csproj", "{0EFF2E29-829B-4B1C-A7C6-47D63B05F52B}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Divergent.Shipping.API.Host", "Divergent.Shipping.API.Host\Divergent.Shipping.API.Host.csproj", "{0059E3AC-0AFA-43F6-B631-200CBB723F94}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ITOps.ViewModelComposition.Json", "ITOps.ViewModelComposition.Json\ITOps.ViewModelComposition.Json.csproj", "{4951C5A4-59D4-4AE1-B74C-B6F667544211}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ITOps.ViewModelComposition.Mvc", "ITOps.ViewModelComposition.Mvc\ITOps.ViewModelComposition.Mvc.csproj", "{536CBB25-3800-4ED5-9A0A-E3FC8FE180A0}"
+EndProject
```
adamralph: I believe this project should be removed from this solution.

demos/ASP.Net Core/ITOps.ViewModelComposition.Gateway/RouteBuilderExtentions.cs
```diff
@@ -0,0 +1,32 @@
+﻿using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
+using Microsoft.AspNetCore.Routing.Constraints;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace ITOps.ViewModelComposition.Gateway
+{
+    public static class RouteBuilderExtentions
+    {
+        public static IRouteBuilder MapComposableGet(this IRouteBuilder routeBuilder,
```
adamralph: Minor nitpick: this parameter should also be on its own line.

demos/ASP.Net Core/ITOps.ViewModelComposition.Gateway/ComposableRouteHandler.cs
```diff
@@ -0,0 +1,46 @@
...
+{
+    class ComposableRouteHandler
+    {
+        public static async Task HandleGetRequest(HttpContext context)
+        {
+            var result = await CompositionHandler.HandleGetRequest(context);
+            if (result.StatusCode == StatusCodes.Status200OK)
+            {
+                string json = JsonConvert.SerializeObject(result.ViewModel, GetSettings(context));
+                context.Response.ContentType = "application/json; charset=utf-8";
```
adamralph: How do we know that we can return JSON? Do we respect the HTTP `Accept` header anywhere in the pipeline?

demos/ASP.Net Core/ITOps.ViewModelComposition.Gateway/ComposableRouteHandler.cs
```diff
@@ -0,0 +1,46 @@
...
+            }
+            else
+            {
+                context.Response.StatusCode = result.StatusCode;
+            }
+        }
+
+        static JsonSerializerSettings GetSettings(HttpContext context)
+        {
+            if (!context.Request.Headers.TryGetValue("Accept-Casing", out StringValues casing))
```
adamralph: Is `Accept-Casing` a standard HTTP header?

mauroservienti: No, I invented it. Is there a special naming convention for custom headers

demos/ASP.Net Core/Divergent.CompositionGateway/Properties/launchSettings.json
```diff
@@ -0,0 +1,27 @@
...
+        "Divergent.CompositionGateway": {
+            "commandName": "Project",
+            "launchBrowser": true,
+            "environmentVariables": {
+                "ASPNETCORE_ENVIRONMENT": "Development"
+            },
+            "applicationUrl": "http://localhost:4458"
+        }
+    }
+}
```
mauroservienti: it's managed by VS when you touch the project property page, I'll fix it, I'm not sure it won't be changed once more.

demos/ASP.Net Core/Divergent.Frontend/Controllers/api/ValuesController.cs
```diff
@@ -0,0 +1,42 @@
...
+        {
+        }
+        
+        // DELETE: api/ApiWithActions/5
+        [HttpDelete("{id}")]
+        public void Delete(int id)
+        {
+        }
+    }
+}
```
mauroservienti: it's just there to prove that we don't break existing functionalities by injecting ourselves into the routing pipeline. We can safely remove it.

demos/ASP.Net Core/Divergent.Frontend/Controllers/HomeController.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Microsoft.AspNetCore.Mvc;
+
+namespace Divergent.Frontend.Controllers
+{
+    public class HomeController : Controller
+    {
+        public IActionResult Index()
+        {
+            return RedirectToRoute(new { controller = "Orders" });
+        }
```
mauroservienti: yes, but that's the default VS template, what's the advantage of making an empty controller looks different from controllers the user will create?

demos/ASP.Net Core/Divergent.Frontend/Controllers/OrdersController.cs
```diff
@@ -0,0 +1,17 @@
...
+        {
+            return View();
+        }
+
+        public IActionResult Details(int id)
+        {
+            return View();
+        }
+    }
+}
```
mauroservienti: As above:

> yes, but that's the default VS template, what's the advantage of making an empty controller looks different from controllers the user will create?

demos/ASP.Net Core/ITOps.ViewModelComposition/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+
+namespace ITOps.ViewModelComposition
+{
+    public static class ServiceCollectionExtensions
+    {
+        public static void AddViewModelComposition(this IServiceCollection services, string assemblySearchPattern = "*ViewModelComposition*.dll")
```
mauroservienti: the parameter is there to allow people start playing with samples and not being forced to use our convention to name assemblies. I'll make it an overload.

demos/ASP.Net Core/ITOps.ViewModelComposition/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+
+            var types = new List<Type>();
+            foreach (var fileName in fileNames)
+            {
+                var temp = AssemblyLoader.Load(fileName)
+                    .GetTypes()
+                    .Where(t =>
+                    {
+                        var typeInfo = t.GetTypeInfo();
+                        return !typeInfo.IsInterface
```
mauroservienti: it's a `!`, we're looking for types that are neither interfaces or abstract classes. How would you change it to make sure we'll pick only concrete types?

demos/ASP.Net Core/ITOps.ViewModelComposition.Gateway/ApplicationBuilderExtensions.cs
mauroservienti: Do you mean like this?

```
public static void RunCompositionGatewayWithDefaultRoutes(this IApplicationBuilder app) => app.RunCompositionGateway(routes =>
{
    routes.MapComposableGet(template: "{controller}/{id:int?}");
    routes.MapRoute("*", context =>
    {
        context.Response.StatusCode = StatusCodes.Status404NotFound;
        return Task.CompletedTask;
    });
});
```

Honestly I can't read it

demos/ASP.Net Core/ASP.Net Core API Gateway.sln
mauroservienti: yes

demos/ASP.Net Core/ITOps.ViewModelComposition.Gateway/ComposableRouteHandler.cs
mauroservienti: no, in this sample we don't respect `Accept` and we have no support for response formatters. It's just an assumption for the sample.
Do you think we should?

mauroservienti:
re the "Solution and project naming" being the one that worked on it for several weeks to me as it is now makes a lot of sense, just because I know what's going on. All your suggestions make sense. Pick the one you prefer and makes most sense from the point of view of someone that never saw this before and I'll adopt that.

> What is the rationale behind the split of types between `ITOps.ViewModelComposition.csproj` and `ITOps.ViewModelComposition.Gateway`? I see that `ITOps.ViewModelComposition.csproj` even has some types in the `ITOps.ViewModelComposition.Gateway`. Is that by design?

`ITOps.ViewModelComposition` is the engine, it has no knowledge at all of the presentation technology. `ITOps.ViewModelComposition.Gateway` knows how to setup a reverse proxy with ASP.Net Core, and `ITOps.ViewModelComposition.Mvc` know how to deal with ASP.Net MVC Core.

> ITOps.ViewModelComposition.Json.csproj seems to contain only some very generic serialization methods which take an HttpContent, assume that it contains JSON, and then asynchronously deserialize that JSON into an ExpandoObject or ExpandoObject[]. Is this specific to view model composition in any way, or should ViewModelComposition be removed from the project name?

it can be called `JsonHelper` but I prefer to have a finger cut than calling an assembly *Helper 😄 
Theoretically it could be merged into `ITOps.ViewModelComposition`, I split it to avoid making `ITOps.ViewModelComposition` dependant on Json.Net, this is it.

demos/ASP.Net Core/Divergent.Frontend/Controllers/HomeController.cs
adamralph: image

Fair enough. Let's leave as-is. The VS templates should really be updated.

demos/ASP.Net Core/ITOps.ViewModelComposition/ServiceCollectionExtensions.cs
adamralph: image

This should do it:

```c#
return !t.GetTypeInfo().IsAbstract && typeof(IRouteInterceptor).IsAssignableFrom(t);
```

If the type is an interface, then `IsAbstract` will return true, negating the first condition. Checking `!IsInterface` is redundant.

demos/ASP.Net Core/ITOps.ViewModelComposition.Gateway/ApplicationBuilderExtensions.cs
adamralph: I was thinking less intrusive, like this:

```c#
public static void RunCompositionGatewayWithDefaultRoutes(this IApplicationBuilder app) =>
    app.RunCompositionGateway(routes =>
    {
        routes.MapComposableGet(template: "{controller}/{id:int?}");
        routes.MapRoute("*", context =>
         {
             context.Response.StatusCode = StatusCodes.Status404NotFound;
             return Task.CompletedTask;
         });
    });
```

I.e. the braces disappear and `=>` is appended to the declaration. Everything else stays the same, even the method body indentation.

BTW - I just noticed that there is an extra space before each of the braces that surrounds the handler delegate that's passed to the `MapRoute` method.

demos/ASP.Net Core/ITOps.ViewModelComposition.Gateway/ComposableRouteHandler.cs
adamralph: image

OK, I think for the sample it's good enough.

Perhaps we should add a comment here to avoid an attendee asking the same question? Something like:

```c#
// For the purposes of the sample, we're not respecting the HTTP Accept header and assuming that a JSON response is OK.
```

demos/ASP.Net Core/ITOps.ViewModelComposition.Gateway/ComposableRouteHandler.cs
adamralph: The convention used to be an `X-`,  prefix, but that was deprecated some time ago.

However, is this overkill? We're already ignoring the Accept header and assuming JSON is OK. Shall we simplify and just assume that a specific casing is also OK?

demos/ASP.Net Core/ITOps.ViewModelComposition.Gateway/RouteBuilderExtentions.cs
```diff
@@ -0,0 +1,33 @@
+﻿using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
+using Microsoft.AspNetCore.Routing.Constraints;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace ITOps.ViewModelComposition.Gateway
+{
+    public static class RouteBuilderExtentions
```
adamralph: Typo: "Extentions" -> "Extensions"

demos/ASP.Net Core/ITOps.ViewModelComposition.Gateway/ComposableRouteHandler.cs
mauroservienti: that was introduced to solve one issue:

* if client is JS you expect camel casing
* if client is something else, e.g. MVC, you expect pascal casing

And in demos we have the same gateway used by JS and MVC

demos/ASP.Net Core/Divergent.Frontend/Controllers/api/ValuesController.cs
mauroservienti: removed

mauroservienti:
> Regarding the solution and project naming, the one that makes the most sense to me is
>
> - `Divergent.WebApi.sln`
>   - `Divergent.WebApi.csproj`

the thing is that it's not WebAPI, WebAPI is already an established term that refers, when we talk in the context of .NET, to ASP.Net WebAPI.

It is a `CompositionGateway` or `APIGateway` or `CompositionAPIGateway`. Because it not _only_ an API Gateway, it's an API Gateway with composition support.

> - `Divergent.Website.sln`
>   - `Divergent.Website.csproj`
> 

OK.

> Thanks for the explanation regarding the ITOps projects. One thing I still don't understand is why `ITOps.ViewModelComposition.csproj` has types in a `ITOps.ViewModelComposition.Gateway` namespace, whereas `ITOps.ViewModelComposition.Gateway.csproj` also exists. Is it that the `ITOps.ViewModelComposition.Gateway` types in `ITOps.ViewModelComposition.csproj` are generic gateway components, whereas the types in `ITOps.ViewModelComposition.Gateway.csproj` are the ASP.NET Core specific implementations?
> 

I don't understand what you mean by

> `ITOps.ViewModelComposition.csproj` has types in a `ITOps.ViewModelComposition.Gateway` 

In `ITOps.ViewModelComposition.Gateway` there are just 3 classes that strictly related to how a reverse proxy works in .NET Core, they act as a bridge allowing the composition engine to be hosted in a ASP.Net Core reverse proxy. The composition engine is hosting aware, it just composes.

> Regarding `ITOps.ViewModelComposition.Json.csproj`, how about renaming it to `ITOps.Json.csproj`?

OK

demos/ASP.Net Core/ITOps.ViewModelComposition.Gateway/ComposableRouteHandler.cs
mauroservienti: Thinking more about it, in theory we could entirely remove it. I'll spike it in another branch.

mauroservienti:
I should have addressed all the comments @adamralph, please take a look.

mauroservienti:
rebased on master

demos/asp-net-core/Divergent.Sales.ViewModelComposition/OrderDetailsViewModelAppender.cs
```diff
@@ -0,0 +1,39 @@
...
+            return HttpMethods.IsGet(verb)
+                && controller.ToLowerInvariant() == "orders"
+                && routeData.Values.ContainsKey("id");
+        }
+
+        public async Task Append(dynamic vm, RouteData routeData, IQueryCollection query)
+        {
+            var id = (string)routeData.Values["id"];
+
+            var url = $"http://localhost:20295/api/orders/{id}";
```
adamralph: It feels a bit weak having the host and port hard coded in here, since the values depend on deployment, and what is in the app config for the host. Is there anyway we can configure this for the hosts that invoke this assembly?

The same comment applies in a few other places.

demos/asp-net-core/Divergent.Sales.ViewModelComposition/OrdersListViewModelAppender.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        IDictionary<dynamic, dynamic> MapToDictionary(dynamic[] orders)
+        {
+            var ordersViewModel = new Dictionary<dynamic, dynamic>();
+
+            foreach (dynamic order in orders)
+            {
+                dynamic vm = new ExpandoObject();
+                vm.OrderNumber = order.Number;
+                vm.OrderItemsCount = order.ItemsCount;
```
adamralph: We can use an initialisier to set these properties.

demos/asp-net-core/Divergent.Shipping.API.Host/Controllers/ShippingInfoController.cs
```diff
@@ -0,0 +1,42 @@
...
+            }
+        }
+
+        [HttpGet]
+        [Route("orders")]
+        public IEnumerable<ShippingInfo> Orders(string ids)
+        {
+            using (var db = new ShippingContext())
+            {
+                var _ids = ids.Split(",".ToCharArray(), StringSplitOptions.RemoveEmptyEntries).Select(s=>int.Parse(s)).ToArray();
```
adamralph: Apologies for this, but `_ids` is setting off my OCD. The underscore convention is usually used for fields (although my preference is `this.`). Perhaps `idArray`?

demos/asp-net-core/Divergent.Sales.ViewModelComposition/OrdersListViewModelAppender.cs
mauroservienti: I'd love to, an initializer would look like:

``` 
dynamic vm = new ExpandoObject()
{
   //set props here
};
```

but compiler complains because `ExpandoObject` doesn't have those properties we're trying to set.

demos/asp-net-core/Divergent.Sales.ViewModelComposition/OrderDetailsViewModelAppender.cs
mauroservienti: it's super hard for a sample, I thought about it, the thing is the way .NET Core allows to inject (via DI) configuration options is not really nice and it might complicate a lot the sample. What if we accept this as is for now, raise an issue to track it and spike something later? so not to complicate even more this PR

demos/asp-net-core/Divergent.Sales.API.Host/Startup.cs
```diff
@@ -0,0 +1,35 @@
...
+            config.MapHttpAttributeRoutes();
+
+            config.Routes.MapHttpRoute(
+                name: "DefaultApi",
+                routeTemplate: "api/{controller}/{id}",
+                defaults: new { id = RouteParameter.Optional }
+            );
+
+            appBuilder.UseCors(CorsOptions.AllowAll);
+            appBuilder.UseWebApi(config);
```
mauroservienti: API Hosts are still full framework based, this is because in the original demo these hosts were an exact copy of the ones used in exercises, it's easier for attendees to connect the dots.
I propose that we leave them unchanged so that they are still as the exercises one and raise an issue to convert them to .NET Core, maybe in the same timeframe we convert exercises to .NET Core.

demos/asp-net-core/ITOps.ViewModelComposition/IViewModel.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.Threading.Tasks;
+
+namespace ITOps.ViewModelComposition
+{
+    public interface IViewModel
```
mauroservienti: It has no real usage, it use Udi that wanted it to mark that a dynamic ViewModel is a `IViewModel`, the idea was that appenders could have had their own interface inheriting from `IViewModel` to express the intent of using a concrete type instead of `dynamic`. I'm 👍 to drop it since we're not using it at all

demos/asp-net-core/ITOps.ViewModelComposition/ISubscription.cs
```diff
@@ -0,0 +1,11 @@
+﻿using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
+using System.Threading.Tasks;
+
+namespace ITOps.ViewModelComposition
+{
+    interface ISubscription
```
mauroservienti: you wanted it 😄 , it was an abstract class before. In any case a non generic type is required here https://github.com/Particular/Workshop.Microservices/blob/asp-net-core-api-gateway/demos/asp-net-core/ITOps.ViewModelComposition/DynamicViewModel.cs#L14 where we need a list of non generic types and generic open types cannot be used (unfortunately)

demos/asp-net-core/ITOps.ViewModelComposition/Subscription-generic.cs
```diff
@@ -0,0 +1,19 @@
+﻿using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
+using System;
+using System.Threading.Tasks;
+
+namespace ITOps.ViewModelComposition
+{
+    class Subscription<T> : ISubscription
+    {
+        private Func<dynamic, T, RouteData, IQueryCollection, Task> subscription;
```
mauroservienti: or maybe `handler` since that's what it is.

demos/asp-net-core/ITOps.ViewModelComposition/Subscription-generic.cs
```diff
@@ -0,0 +1,19 @@
+﻿using Microsoft.AspNetCore.Http;
```
mauroservienti: we already had this conversation a few weeks ago 😄 at that time there was a `Subscription` class, that is now the `ISubscription` interface. `Subscription`1.cs` is not the convention, other than being horrible and hard to type on on en-US keyboards, is the way decompilers do that.

I can rename it to `Subscription.cs`

demos/asp-net-core/ITOps.ViewModelComposition/IRouteInterceptor.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Threading.Tasks;
+
+namespace ITOps.ViewModelComposition
+{
+    public interface IRouteInterceptor
+    {
+        bool Matches(RouteData routeData, string httpVerb);
+    }
+
+    public interface ISubscribeToCompositionEvents : IRouteInterceptor
```
mauroservienti: oh yes, I like it as well. I missed this one, these files were created a few months ago with Udi on call, live typing in VS. I'll split them.

demos/asp-net-core/Divergent.CompositionGateway.sln
```diff
@@ -0,0 +1,100 @@
...
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Divergent.Sales.ViewModelComposition.Events", "Divergent.Sales.ViewModelComposition.Events\Divergent.Sales.ViewModelComposition.Events.csproj", "{5875B01D-5A47-4B08-99B2-48D217F09B30}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Shipping", "Shipping", "{9D36D6A4-373C-4CB0-8B54-D15505E2AE91}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Divergent.Shipping.ViewModelComposition", "Divergent.Shipping.ViewModelComposition\Divergent.Shipping.ViewModelComposition.csproj", "{657A5136-DDDE-493F-9247-B2789B3294F3}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Divergent.Sales.Data", "Divergent.Sales.Data\Divergent.Sales.Data.csproj", "{80F82F7D-3F0A-4316-87CD-49F19DBD5B85}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Divergent.Sales.API.Host", "Divergent.Sales.API.Host\Divergent.Sales.API.Host.csproj", "{DC8ADA37-837C-490F-BDFC-FC8D1001417E}"
```
mauroservienti: As said they are the copy of the endpoint used by the exercises. So 👎 to rename them unless we fix them all. Nitpick, API is an acronym thus conventions say it should be uppercase

demos/asp-net-core/Divergent.CompositionGateway.sln
```diff
@@ -0,0 +1,100 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 15
+VisualStudioVersion = 15.0.26430.16
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Divergent.CompositionGateway", "Divergent.CompositionGateway\Divergent.CompositionGateway.csproj", "{C0E4CD4F-657A-4872-969C-D97768636BDA}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "IT-Ops", "IT-Ops", "{DB3EFAF6-6B45-48ED-BCB9-5F44FAB30CF9}"
```
mauroservienti: All over the place (in slides) we `IT/Ops`, with the `/`, that cannot be used in file names. I'll align it with project names.

demos/asp-net-core/ITOps.ViewModelComposition.Gateway/RouteBuilderExtentions.cs
```diff
@@ -0,0 +1,33 @@
+﻿using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
+using Microsoft.AspNetCore.Routing.Constraints;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace ITOps.ViewModelComposition.Gateway
+{
+    public static class RouteBuilderExtensions
```
mauroservienti: Ooops, missed that

mauroservienti:
> Somewhat related to the last point, in ITOps.ViewModelComposition.csproj, there is only one type in each of the namespaces ITOps.ViewModelComposition.Engine and ITOps.ViewModelComposition.Gateway. Should we collapse those two types into the ITOps.ViewModelComposition namespace?

Yeah, we should. I believe I noticed another couple of things that are in the wrong namespace, I'll do another pass.

> Did you get anywhere with the JSON casing spike?

No, not yet. It affects a lot of projects in this demo, to keep changes to this PR in scope I propose we revisit as we have this merged. It's something that we can drop but it's also transparent to the logic of the sample, the problem is that changes will be pervasive to all projects, only complicating this PR.

demos/asp-net-core/Divergent.Shipping.API.Host/Controllers/ShippingInfoController.cs
mauroservienti: As above these are copies (or were originally copies) of the APIs used in the exercise. I can change it.

mauroservienti:
>>In ITOps.ViewModelComposition.Gateway there are just 3 classes that strictly related to how a reverse proxy works in .NET Core, they act as a bridge allowing the composition engine to be hosted in a ASP.Net Core reverse proxy. The composition engine is hosting aware, it just composes.

>That tells me that ITOps.ViewModelComposition.Gateway.csproj contains the ASP.NET Core-specific implementation of abstractions defined in ITOps.ViewModelComposition.csproj. Is that correct? In that case, I would consider renaming ITOps.ViewModelComposition.Gateway.csproj to ITOps.ViewModelComposition.AspNetCore.

Originally `ITOps.ViewModelComposition.Gateway` was collapsed into `ITOps.ViewModelComposition`, but then whenever you're using `MVC` you have available at configuration time all the `Gateway` configuration extensions that for MVC make no sense, that's the main reason I split the `Gateway` into its own project, so that its purpose was clear. We can easily merge `ITOps.ViewModelComposition.Gateway` back into `ITOps.ViewModelComposition`.

>Although, I'm unsure about that, since I see that ITOps.ViewModelComposition.csproj also has a dependency on Microsoft.AspNetCore.Routing.nupkg.

`ITOps.ViewModelComposition` (the engine) depends on HTTP only, the routing dependency is a convenient way to have all the facilities that the routing provide to analyze HTTP requests, otherwise we would have needed to reinvent the wheel. Luckily routing is not anymore double linked to MVC as it was before, and now is part of the core part of AspNetCore, and the engine is designed for AspNetCore only.

demos/asp-net-core/ITOps.ViewModelComposition/CompositionHandler.cs
```diff
@@ -0,0 +1,57 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace ITOps.ViewModelComposition.Gateway
+{
+    public class CompositionHandler
+    {
+        public static async Task<(dynamic ViewModel, int StatusCode)> HandleGetRequest(HttpContext context)
```
mauroservienti: It's used by the `CompositionActionFilter` as well, in the MVC project. This is the main reason I was reluctant to split the solution and make 2 with shared projects, it's impossible to see the whole picture.

demos/asp-net-core/ITOps.ViewModelComposition/AssemblyLoader.cs
```diff
@@ -0,0 +1,26 @@
...
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Runtime.Loader;
+
+namespace ITOps.ViewModelComposition
+{
+    public static class AssemblyLoader
+    {
+        public static Assembly Load(string assemblyFullPath)
```
mauroservienti: yes and no, it's a generic way to load assemblies in .NET Core not connected to what we do here, it should go in its own "helper" assembly. It is required if we want (as I did for the DDD SouthWest demo) to introduce ViewComponents that are a very interesting Mvc Core feature for UI Composition

mauroservienti:
@adamralph addressed most of the changes, and replied to a couple of comments.

demos/asp-net-core/Divergent.CompositionGateway.sln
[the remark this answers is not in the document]
adamralph: The [.NET capitalisation conventions](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions) state PascalCasing for any acronym over two letters in length (although the BCL violates that in numerous places). However, it's only a nitpick and I agree that we should fix everything or nothing, so let's leave this as-is for now.

demos/asp-net-core/ITOps.ViewModelComposition/Subscription.cs
```diff
@@ -0,0 +1,19 @@
+﻿using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
+using System;
+using System.Threading.Tasks;
+
+namespace ITOps.ViewModelComposition
+{
+    class Subscription<T> : ISubscription
+    {
+        private Func<dynamic, T, RouteData, IQueryCollection, Task> subscription;
```
adamralph: image

@mauroservienti did you intend to rename this?

Personally I like using a verb for the name of a delegate, since then invocation is of the form `handle()`, but I won't oppose `handler` if that's what you prefer.

demos/asp-net-core/Divergent.Sales.Data/Models/Order.cs
```diff
@@ -0,0 +1,21 @@
...
+    public class Order
+    {
+        public int Id { get; set; }
+
+        public DateTime DateTimeUtc { get; set; }
+
+        public ICollection<Item> Items { get; set; }
+    }
+
+    public class Item
```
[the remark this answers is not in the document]
adamralph: Should we split these types into separate files?

demos/asp-net-core/ITOps.ViewModelComposition/DynamicViewModel.cs
```diff
@@ -0,0 +1,83 @@
+﻿using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
+using System;
+using System.Collections.Generic;
+using System.Dynamic;
+using System.Threading.Tasks;
+
+namespace ITOps.ViewModelComposition.Engine
```
adamralph: As per https://github.com/Particular/Workshop.Microservices/pull/59#issuecomment-318675121, should this be moved to `ITOps.ViewModelComposition`?

demos/asp-net-core/ITOps.ViewModelComposition/CompositionHandler.cs
```diff
@@ -0,0 +1,57 @@
+﻿using ITOps.ViewModelComposition.Engine;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
+using Microsoft.Extensions.DependencyInjection;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace ITOps.ViewModelComposition.Gateway
```
adamralph: As per https://github.com/Particular/Workshop.Microservices/pull/59#issuecomment-318675121, should this be moved to `ITOps.ViewModelComposition`?

demos/asp-net-core/Divergent.Website.sln
```diff
@@ -0,0 +1,107 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 15
+VisualStudioVersion = 15.0.26430.16
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "ITOps", "ITOps", "{DB3EFAF6-6B45-48ED-BCB9-5F44FAB30CF9}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ITOps.ViewModelComposition.Gateway", "ITOps.ViewModelComposition.Gateway\ITOps.ViewModelComposition.Gateway.csproj", "{1B2BC5FF-3F95-4429-8AE1-5AC6C99B689A}"
```
adamralph: Should this project be removed from the Website solution?

demos/asp-net-core/ITOps.ViewModelComposition/Subscription.cs
mauroservienti: done

demos/asp-net-core/ITOps.ViewModelComposition/DynamicViewModel.cs
mauroservienti: done

demos/asp-net-core/ITOps.ViewModelComposition/CompositionHandler.cs
mauroservienti: done

demos/asp-net-core/Divergent.Sales.Data/Models/Order.cs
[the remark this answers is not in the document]
mauroservienti: @adamralph 

> Should we split these types into separate files?

yes and no. The back-end projects (WebAPI) used by this demo, right now are an exact copy of the back-end projects used by exercises. They all should be changed probably to .NET Core as soon as NServiceBus on .NET Core is out of beta, till then I prefer not to diverge. It's indeed not a big change, but still I prefer not to diverge. Can we postpone this refactoring to a later time?

mauroservienti:
All comments addressed.

mauroservienti:
should this be upgraded to `netcoreapp2.0`?

demos/asp-net-core/Divergent.Sales.Data/Models/Order.cs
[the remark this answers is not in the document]
adamralph: 👍 fine by me.

mauroservienti:
squashed into 1 single commit @adamralph. And yes, `netcoreapp2.0` can be handled as a separate PR

mauroservienti:
🎆 🎉

PR #60, opened 2017-04-16
Ensure subscribers can subscribe before appenders are executed

mauroservienti:
Depending on the IoC container of choice interceptors are returned in a non deterministic order. It might happen that appenders are invoked before subscribers have the opportunity to subscribe to events, causing the entire composition process to fail.

@dvdstelt please review.


mauroservienti:
I'm self merging this as it is a potential bug, technically due to the way Windsor resolves dependencies the but won't surface ever in this specific sample. If the IoC container is changed the sample might immediately stop working.

PR #61, opened 2017-04-16
Dependencies refactoring

mauroservienti:
After demoing UI Composition to @adamralph and @DavidBoike I realized that many MVC Demo dependencies are there simply because they were needed. The demo evolved a lot from the very beginning and many of them are not needed anymore. This PR refactors the whole demo removing all what is not strictly needed for the demo itself, so to keep as less magic as needed and as simple as possible.

/cc @dvdstelt 

## PoA

* [x] removes nuget packages: drops all not anymore needed packages
* [x] DynamicViewModel does not depend anymore on in memory MessageBroker: originally the dynamic view model needs in term of event dispatching were much more complex, now a full in memory bus with support for events inheritance and complex subscriptions logic is not needed anymore
* [x] Removes in memory broker dependency from composition filter: due to previous one
* [x] removes dependency on WindsorBootstrapper: compared to the beginning usage of IoC in the frontend is now limited to the composition steps and MVC controllers dont't need anymore to be resolved via the container, making IoC less compelling
* [x] Explicit filter registration: it's simpler to explicit register the composition filer instead of relying on registering all `IResultFilter` implementation
* [x] WindsorBootstrapper dependency in WebAPI Hosts: it was due to the way demo originally started, but it is not uned anymore in the codebase except that is configured at startup, it's basically useless.


demos/CompositeUI-MVC/Divergent.Frontend/App_Start/ContainerConfig.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        public static IWindsorContainer ConfigContainer()
+        {
+            var container = new WindsorContainer();
+            container.Kernel.Resolver.AddSubResolver(new ArrayResolver(container.Kernel, true));
+
+            var bin = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "bin");
+            var concreteTypes = Directory.EnumerateFiles(bin, "Divergent*.dll")
+                .SelectMany(path => Assembly.LoadFile(path).GetTypes())
+                .Where(t => !t.IsInterface);
```
DavidBoike: `!t.IsAbstract` ?

demos/CompositeUI-MVC/Divergent.ITOps.ViewModelComposition/DynamicViewModel.cs
```diff
@@ -77,9 +84,28 @@ public override bool TryInvokeMember(InvokeMemberBinder binder, object[] args, o
             return false;
         }
 
+        public override IEnumerable<string> GetDynamicMemberNames()
+        {
+            return base.GetDynamicMemberNames()
+                .Union(properties.Keys)
+                .Union(new[] { "RaiseEventAsync", "RaiseEvent" });
```
DavidBoike: In a real app, would this be a hot path? Would using `yield return` be better than relying upon LINQ?

mauroservienti: agree. In this specific sample that method in never called. Anyway do you mean something like this:

```
foreach (var item in base.GetDynamicMemberNames())
{
      yield return item;
}

foreach (var item in properties.Keys)
{
      yield return item;
}

yield return "RaiseEventAsync";
```

DavidBoike: You forgot `yield return "RaiseEvent";` but yes. Unless it would be normal for there to be overlap that would cause an issue? Then you could probably maintain your own HashSet cheaper than calling Union twice. (Although that's a guess.)

mauroservienti: yes, I wrote it in the comment. I could even be simpler:

```
foreach (var item in properties.Keys)
{
      yield return item;
}

yield return "RaiseEventAsync";
yield return "RaiseEvent";
```

dropping the `base` call, since base class has no members at all in such a scenario, never.

dvdstelt:
Most of it seems legit, but why is this still using SQLite?

mauroservienti:
so far there are no needs, for demos, to use `LocalDb`, or anything other than `SQLite`.

In demos we're just reading data, not writing anything and there is no NServiceBus at all. I prefer to avoid changing it unless it is strictly required.

@dvdstelt thoughts?

dvdstelt:
@mauroservienti If we're just reading, why not have static dictionaries? Way easier and 100% sure it'll work.

mauroservienti:
I agree, but again should we change it now? It has been like this from the very beginning.

mauroservienti:
rebased on master after mergin #60

demos/CompositeUI-MVC/Divergent.Frontend/App_Start/ContainerConfig.cs
```diff
@@ -0,0 +1,51 @@
+﻿using Castle.MicroKernel.Registration;
+using Castle.MicroKernel.Resolvers.SpecializedResolvers;
+using Castle.Windsor;
+using Divergent.ITOps.ViewModelComposition;
+using System;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Web;
```
adamralph: This line is redundant.

demos/CompositeUI-MVC/Divergent.Frontend/App_Start/ContainerConfig.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.Web.Mvc;
+using System.Web.Optimization;
```
adamralph: This line is redundant.

demos/CompositeUI-MVC/Divergent.Frontend/App_Start/ContainerConfig.cs
```diff
@@ -0,0 +1,51 @@
+﻿using Castle.MicroKernel.Registration;
```
adamralph: Should we be sorting the namespaces with `System.*` at the top?

demos/CompositeUI-MVC/Divergent.Frontend/App_Start/ContainerConfig.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Web;
+using System.Web.Mvc;
+using System.Web.Optimization;
+
+namespace Divergent.Frontend
+{
+    public class ContainerConfig
```
adamralph: This class should be static.

demos/CompositeUI-MVC/Divergent.Frontend/App_Start/ContainerConfig.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        public static IWindsorContainer ConfigContainer()
```
adamralph: It feels like the class should be called `ContainerFactory` and this method should be called `Create()`.

demos/CompositeUI-MVC/Divergent.Frontend/App_Start/ContainerConfig.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            var container = new WindsorContainer();
+            container.Kernel.Resolver.AddSubResolver(new ArrayResolver(container.Kernel, true));
+
+            var bin = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "bin");
+            var concreteTypes = Directory.EnumerateFiles(bin, "Divergent*.dll")
+                .SelectMany(path => Assembly.LoadFile(path).GetTypes())
+                .Where(t => !t.IsInterface && !t.IsAbstract);
```
adamralph: The `concreteTypes` variable (and it dependency `bin`) are not used until the `foreach` block, it would be nicer if these two assignments could be moved below the next `Register` call, so that they sit next to the `foreach` block.

demos/CompositeUI-MVC/Divergent.Frontend/App_Start/ContainerConfig.cs
```diff
@@ -0,0 +1,51 @@
...
+
+            container.Register(Component.For<IResultFilter>()
+                    .ImplementedBy<CompositionActionFilter>()
+                    .LifestyleTransient());
+
+            foreach (var t in concreteTypes.Where(t => typeof(IRouteInterceptor).IsAssignableFrom(t)))
```
adamralph: Can we rename `t` to `type`?

demos/CompositeUI-MVC/Divergent.ITOps.ViewModelComposition/DynamicViewModel.cs
```diff
@@ -2,50 +2,57 @@
...
 using System.Dynamic;
 using System.Linq;
-using System.Text;
 using System.Threading.Tasks;
-using Topics.Radical.ComponentModel.Messaging;
 
 namespace Divergent.ITOps.ViewModelComposition
 {
-    class DynamicViewModel : DynamicObject, ISubscriptionStorage, IViewModel
+    abstract class Subscription
```
adamralph: Can we put this class in its own file?

demos/CompositeUI-MVC/Divergent.ITOps.ViewModelComposition/DynamicViewModel.cs
```diff
@@ -2,50 +2,57 @@
...
+    {
+        public abstract Task Invoke(dynamic viewModel, object @event, RequestContext requestContext);
+    }
+
+    class Subscription<T> : Subscription
```
adamralph: Can we put this class in its own file?

demos/CompositeUI-MVC/Divergent.ITOps.ViewModelComposition/DynamicViewModel.cs
```diff
@@ -2,50 +2,57 @@
 using System.Collections.Generic;
 using System.Dynamic;
 using System.Linq;
```
adamralph: This line is redundant.

demos/CompositeUI-MVC/Divergent.ITOps.ViewModelComposition/DynamicViewModel.cs
```diff
@@ -2,50 +2,57 @@
...
+        {
+            this.subscription = subscription;
+        }
 
-        public DynamicViewModel(IMessageBroker inMemoryBroker, RequestContext request)
+        public override Task Invoke(dynamic viewModel, object @event, RequestContext requestContext)
         {
-            this.inMemoryBroker = inMemoryBroker;
-            this.request = request;
+            return subscription(viewModel, (T)@event, requestContext);
```
adamralph: This can be an expression-bodied member.

demos/CompositeUI-MVC/Divergent.Shipping.API.Host/Startup.cs
```diff
@@ -1,16 +1,8 @@
-﻿using Castle.MicroKernel.Registration;
-using Divergent.Shipping.Data.Context;
-using Microsoft.Owin.Cors;
+﻿using Microsoft.Owin.Cors;
 using Newtonsoft.Json.Serialization;
 using Owin;
-using Radical.Bootstrapper;
-using Radical.Bootstrapper.Windsor.WebAPI.Infrastructure;
 using System;
```
adamralph: Missed this in my review: this line is redundant.

adamralph:
Oh, and:

> @mauroservienti If we're just reading, why not have static dictionaries? Way easier and 100% sure it'll work.

I tend to agree with @dvdstelt - the simpler the better.

mauroservienti:
> Why are we committing the .suo file?

I hate it 😄 but:

* we have multiple startup projects that need to be configured. 
* we want users to be able to just press F5 and run (we discussed this many times)

If we can find a better solution I'm all for it 🎉 

> Can we choose a single name to replace the three we currently have in play? Those are:

aligned to `ASP.Net MVC`, dropping demo from the solution name and folder name given that it lives in the `demos` folder

> Root namespace: Divergent

No please, no. changing the root namespace everywhere is a mess. @dvdstelt do you want top do it?

mauroservienti:
>>@mauroservienti If we're just reading, why not have static dictionaries? Way easier and 100% sure it'll work.

> I tend to agree with @dvdstelt - the simpler the better.

the thing is that we have this "secret" plan to add _writes_, that right now is as easy as just write the code, or if we really want also change the connection string to point to a SQL instance. Changing to static data means more work in the near future. So I'm 👎 to move to static data.

demos/CompositeUI-MVC/Divergent.Frontend/App_Start/ContainerConfig.cs
mauroservienti: I prefer strictly alphabetical order

mauroservienti:
addressed all review comments @adamralph

dvdstelt:
> the thing is that we have this "secret" plan to add writes,

Oooh no! Now it's not secret anymore, this is a public repo! 😄 

And the root namespace _is_ divergent, so I'm not getting what needs to be changed?

adamralph:
@dvdstelt see my point about naming in https://github.com/Particular/Workshop.Microservices/pull/61#pullrequestreview-33424219

demos/ASP.Net MVC/Divergent.ITOps.ViewModelComposition/Divergent.ITOps.ViewModelComposition.csproj
```diff
@@ -109,6 +88,8 @@
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="RequestContext.cs" />
     <Compile Include="StringExtensions.cs" />
+    <Compile Include="Subscription %28generic%29.cs" />
```
adamralph: This file name feels a bit awkward, and it doesn't allow for the introduction of a separate file for `Subscription<T, V>`.

I'm not sure there's really an accepted standard for naming the file containing the generic equivalent of an existing non-generic type, but one I've seen used previously is ``Subscription`1.cs``, i.e. matching the "CLR type name".

mauroservienti: Thank you,  bit no, once we'll have the problem of a second file I'll be happy to change the convention. AFAIK the CRL name is when the source is reverse engineered, but I might be wrong.

mauroservienti:
About the `Divergent` root namespace, I'm all for changing it to something like `Particular.Demos` and maybe exercises to `Particular.Exercises`, or similar. It's simply super annoying.
If we can come to a decision soon, maybe today, I can also adjust all the projects in the second demo I'm working on right now.

adamralph:
I'm happy with leaving the namespace as-is for now.

mauroservienti:
As discussed with @adamralph I renamed the `Subscription (generic).cs` file to `Subscription-generic.cs`. This is ready to be merged.

PR #65, opened 2017-06-08
ignore v15

adamralph:
VS2017 respects the VS2015 .suo file, but it then creates its own version of it in a `v15` folder. There's no need for us to commit these files, until we switch to VS2017 as the minimum supported version.


adamralph:
@Particular/microservices-workshop-maintainers please review.

PR #70, opened 2017-06-08
Fixed file location

dvdstelt:
Fixed file location of a single message in the /before/ of exercise 02.


mauroservienti:
This should not be changed, if you look at the after the command is used internally via a SendLocal when Finance needs to kick off the Payment process, it's an internal command sent by Finance to Finance just to decouple from the creation of the order. That why we originally put this Command in the Finance project and not in the messages one

https://github.com/Particular/Workshop.Microservices/blob/master/exercises/src/02%20Publish-Subscribe/after/Divergent.Finance/Handlers/OrderSubmittedHandler.cs

adamralph:
And if this were a legitimate change, it would also have to be done in the "after" project, no? And also in the Sagas and Integration solutions, so six places in total? Right now, this PR only changes one of those six solutions.

mauroservienti:
As agreed with @adamralph we don't want to change the file location in all projects, closing for now

dvdstelt:
> And if this were a legitimate change, it would also have to be done in the "after" project, no? And also in the Sagas and Integration solutions, so six places in total? Right now, this PR only changes one of those six solutions.

I didn't check, my bad. I thought it was a mistake in the solution, because mine wasn't compiling anymore.

PR #90, opened 2017-06-21
missing ConfigureAwait

SimonCropp:


mauroservienti:
Given that `ConfigureAwait` has an effect only on code running in the context of a `SynchronizationContext` hat doesn't exist in ASP.Net Core do you think it's worth adding it?

SimonCropp:
good point

PR #97, opened 2017-07-31
[SPIKE] Simple data-load panel

mauroservienti:
@adamralph this is the simplest form of `data-load` panel that can be add. We could also tweak it to make the loading message _bindable_ from the controller so that in case of error we can show some text in the same panel without adding complex directives or other components.


exercises/01-composite-ui/before/Divergent.Frontend/app/branding/orders/ordersView.html
```diff
@@ -1,7 +1,7 @@
 ﻿<section>
-    Orders
+    <h3>Orders list</h3>
```
adamralph: I think we can drop the word "list".

exercises/01-composite-ui/before/Divergent.Frontend/app/branding/orders/ordersView.html
```diff
@@ -12,4 +12,8 @@
         </li>
     </ul>
 
+    <div ng-show="ctrl.loadingData" class="data-load">
+        Data is loading....
```
adamralph: Perhaps we could have a spinner here instead (with "Loading" alt text)? E.g.

ajax-loader

exercises/01-composite-ui/before/Divergent.Frontend/app/modules/sales/orders/ordersController.js
```diff
@@ -6,13 +6,19 @@
...
                 ctrl.orders = null;
 
-                ctrl.isBusy = backendCompositionService
+                backendCompositionService
                     .get('orders-list', { pageIndex: 0, pageSize: 10 })
                     .then(function (viewModel) {
                         ctrl.orders = viewModel.orders;
+                    })
+                    .catch(function (error) {
+                        $log.error('Ooops, something went wroing: ', error);
```
adamralph: I agree that we should do _something_ in the UI in this case. I'm not sure exactly what's involved in your "bindable" suggestion, but it _sounds_ good. 😉

mauroservienti:
> This approach looks good. Any downsides?

Not that I can think of. I'll get back to this probably next week.

mauroservienti:
I took another direction, now it looks like this:

image

mauroservienti:
And in case of an error a red panel appears

exercises/01-composite-ui/before/Divergent.Frontend/app/branding/orders/ordersView.html
```diff
@@ -1,5 +1,5 @@
-﻿<section>
-    Orders
+﻿<section cg-busy="ctrl.isLoading">
+    <h3>Orders list</h3>
```
adamralph: Can we remove the word "list"?

adamralph: Can we remove the word "list"?

exercises/01-composite-ui/before/Divergent.Frontend/assets/site.css
```diff
@@ -10,3 +10,13 @@ body {
...
+
+.data-load {
+    color: #5e5e5e;
+    height: 50px;
+    padding: 15px 15px;
+    font-size: 18px;
+    line-height: 20px;
+    font-style: italic;
+    text-align: center;
+}
```
adamralph: This class is now redundant and should be removed.

exercises/01-composite-ui/before/Divergent.Frontend/Divergent.Frontend.csproj
```diff
@@ -117,7 +120,9 @@
   <ItemGroup>
     <Compile Include="Properties\AssemblyInfo.cs" />
   </ItemGroup>
-  <ItemGroup />
+  <ItemGroup>
+    <Folder Include="app\directives\" />
```
adamralph: Is this required? The folder is empty.

exercises/01-composite-ui/before/Divergent.Frontend/lib/angular-busy/angular-busy.min.css
```diff
@@ -0,0 +1,308 @@
+﻿.cg-busy {
```
adamralph: This file is named `min` but is not minified. Is it the wrong file?

mauroservienti:
This PR was meant just to show how to add a data loading, I plan to close this one once we are in agreement as it seems we are, and create a new one to implement it on all the exercises.

adamralph:
@mauroservienti OK, understood. I'll rename to a spike then.

And I'm 👍 to go ahead with this approach!

mauroservienti:
Closing as replaced by #101

PR #105, opened 2017-08-09
Exercise 3 - Data loading spinner

mauroservienti:


exercises/03-sagas/before/Divergent.Frontend/app/branding/orders/ordersView.html
```diff
@@ -1,11 +1,10 @@
-﻿<section>
+﻿<section cg-busy="ctrl.isLoading">
     Orders
 
     <div>
         <button ng-click="ctrl.refreshOrders()">Refresh</button>
         <button ng-click="ctrl.createNewOrder()">Create new order</button>
     </div>
-
```
adamralph: A very minor point, but this line was not deleted in the `after` solution. It was kept there, but with the spaces removed. This generates a new diff when I do a directory comparison between before and after, so it would be good to keep a blank line here to avoid that.

mauroservienti: I can't do that till Monday, can you directly fix it and push it?

adamralph: Sure, will do.

PR #109, opened 2017-08-11
[WIP] Add ServiceInsight and ServicePulse

### LINKED ISSUE #89: Add ServiceInsight and ServicePulse
In the last delivery of the workshop at NDC Oslo 2017, I instructed attendees to look at ServiceInsight as a part of the exercises that they were doing, as well as adding the Saga Audit plugin for the saga exercises.

For the existing NServiceBus users/customers who attended, experiencing Pulse and Insight first-hand was very significant. I heard them saying that there was no way that they'd go back to working without them, almost regardless of how much they cost.

In short, I propose that we incorporate more instruction into the workshop around how to install, configure, and use Pulse and Insight in the workshop.

I'd also suggest that this be prioritized so that it can be done in time for the next delivery of the workshop which, I believe, is Jimmy doing it in NDC Sydney.

adamralph:
Fixes #89

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,39 @@ The solutions contain single page applications (SPAs) and use `IIS Express`. To 
   - Ensure all HTML template files are closed when the application is run, or:
   - Manually change the browser address to the root URL.
 
+## Preparing your machine for advanced exercises
+
+If you completed all the standard exercises, first of all, well done!
+
+To complete the advanced exercises, you'll need to install the rest of the Particular Service Platform:
+
+- ServicePulse: a web app which monitors your endpoints.
+- ServiceInsight: a desktop app which visualises the flow of messages within and between endpoints in detail.
+- ServiceControl: a Windows service which collects the data that drives ServicePulse and ServiceInsight.
+
+Run the [Particular Platform Installer](https://particular.net/start-platform-download) again, selecting ServiceControl, ServiceInsight, and ServicePulse.
+
+When installation has completed, click "Start ServiceControl Management".
+
+You may now close the Particular Service Platform Installation window.
+
+In the ServiceControl Management window, click "Add new instance".
+
+In the settings for the new instance, there are two choices you must make:
+
+- "Transport": choose "MSMQ".
+- "Audit forwarding": choose "Off".
+
+All other settings should be left with their default values.
+
+Click "Add".
+
+After the instance has been added, make a note of the name of the instance. This will usually be "Particular.ServiceControl".
+
+You may now close the ServiceControl Management window.
+
+You're now ready to attempt the advanced exercises!
+
 ## FAQ
 
 If the answer to your question is not listed here, consult your on-site trainer.
--- a/exercises/05-platform/instructions.md
+++ b/exercises/05-platform/instructions.md
@@ -1,0 +1,223 @@
+# Exercise 5: Particular Software Platform
+
+**Important: Before attempting the exercise, please ensure you have followed [the instructions for preparing your machine](README.md#preparing-your-machine-for-the-workshop) and that you have read [the instructions for running the exercise solutions](/README.md#running-the-exercise-solutions).**
+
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -103,6 +103,39 @@ The solutions contain single page applications (SPAs) and use `IIS Express`. To
...
+In the settings for the new instance, there are two choices you must make:
+
+- "Transport": choose "MSMQ".
+- "Audit forwarding": choose "Off".
+
+All other settings should be left with their default values.
+
+Click "Add".
+
+After the instance has been added, make a note of the name of the instance. This will usually be "Particular.ServiceControl".
```
mauroservienti: Is this needed? That's the default value, meaning that installing the heartbeat plugin is enough and no further configuration is needed

adamralph: Is it not needed when configuring the saga audit plug in?

mauroservienti: That specifically I don't remember, but it should not

adamralph: From [the docs](https://docs.particular.net/servicecontrol/plugins/saga-audit#configuration-servicecontrol-queue):

> The queue name is based on the ServiceControl instance name...

which means you have to configure it like so:

```C#
var endpointConfiguration = new EndpointConfiguration("myendpoint");
endpointConfiguration.SagaPlugin("Particular.ServiceControl");
```

mauroservienti: Yup. Had a look at the source code, and there isn't anymore a default value

dvdstelt:
Hmmm, wait a minute. I approved this, but the advanced exercises aren't there yet, are they!? :-)
I suggest we add the documentation, but don't create any "before" or "after" folders. If anything goes wrong, this is an excellent chance to read documentation and get hands-on experience. And they are 'advanced' after all, right?

adamralph:
@dvdstelt 

> Hmmm, wait a minute. I approved this, but the advanced exercises aren't there yet, are they!? :-)

Correct, that's why it's marked WIP.

> I suggest we add the documentation, but don't create any "before" or "after" folders.

The idea (as I understand it) is to use the existing solutions (1 to 4) and only to add some instructions for advanced exercises that will involve using SP and SI. I don't think we should alter the solutions at all if we can avoid it. You could argue that we should alter the "after" solution to reflect the state after the advanced exercises have been completed, but I would argue that this isn't required. Those who are moving on to the advanced exercises can be left a little more "on their own", without a reference "after" solution to look at for clues.

/cc @mauroservienti

dvdstelt:
Please review advanced exercise. I added @tmasternak and @SzymonPobiega because they might want to know about it as well, but it is very unlikely this will come up in a 1 day workshop! But you never know, some devs are so extremely fast! ;-)

dvdstelt:
Aaargh, some files are included that certainly should not have been. Let me try to fix.

For the reviewers, it only concerns the new .md file! :)

dvdstelt:
Hmmm, I thought I had already replied to this thread.

- Module 03 already has advanced exercises.
- Module 04 doesn't need them, as this is the advanced exercise
- There is no solution to maintain, They should pick this up after they finish 04. Or worst case scenario they pick the `after` solution of module 04.

I specifically created exercise 05 so people know it's a separate exercise.
- It's totally focused on the platform, not so much on the architectural ideas from the modules presented
- It's for those who finish all their exercises early. But those attendees who can't finish exercise 04 in time, they won't feel bad about not finishing an exercise.
- It's quite large to 'just' include it at the end of module 04.
- People can pick it up at home, if they're really eager to experiment with the exercises.
- We could even point people not doing the workshop to this exercises, for example via support cases or LiveChat.

Module 01 & 02 could perhaps use some advanced exercises as well. But I was thinking more about doing module 01 to 04 as advanced modules and not include 100% of the code to copy-n-paste.

adamralph:
Discussed with @dvdstelt. He will send another PR showing an alternative approach with SI and SP content in advanced exercise sections spread across the current exercises 2,3, and 4. Marking this PR as WIP while we decided between the two approaches.

adamralph:
Closing in favour of #126

PR #120, opened 2017-09-11
Exercise 2 review for TWF

tmasternak:
## General notes
  * It takes quite some time to hit `F5` and see if it runs. Feels like big-bang approach. Should we make the whole process more incremental? e.g. put explicit CTA to run current version of the the exercise?
  * How much commentary do we expect from the Trainer? e.g. 
     * Inside `OrderSubmittedHandler` in `Finance` it might not be obvious why we need the `InitiatePaymentProcessCommand`
      * Should we go through `PaymentClient` to show what kind of data it requires to make a payment before we show `InitiatePaymentProcessCommand` creation?
      * Should we mention potential ordering issues in `Shipping`?
      * Do we go through error cases or wait until `Exercise 3`?

### DIFF
--- a/exercises/02-publish-subscribe/instructions.md
+++ b/exercises/02-publish-subscribe/instructions.md
@@ -141,7 +141,7 @@ routing.RegisterPublisher(typeof(OrderSubmittedEvent), "Divergent.Sales");
 
 ### Step 8
 
-When Finance receives the `OrderSubmittedEvent` message it records prices for the items that belong to the submitted order (this is need to make sure that even when the product prices change the client will be charged the an amount first show on the UI - value returned through `Divergent.Finance.API`). And finally initiate the payment process by sending the `InitiatePaymentProcessCommand` message.
+When Finance receives the `OrderSubmittedEvent` message it records prices for the items that belong to the submitted order (this is needed to make sure that even when the product prices change, the client will be charged the amount first shown in the UI - value returned through `Divergent.Finance.API`). And finally initiate the payment process by sending the `InitiatePaymentProcessCommand` message.
 
 ```c#
 namespace Divergent.Finance.Handlers
--- a/exercises/02-publish-subscribe/instructions.md
+++ b/exercises/02-publish-subscribe/instructions.md
@@ -97,7 +97,7 @@ The `OrderSubmittedHandler` should process the `OrderSubmittedEvent` published b
 
 ### Step 3
 
-In the `Divergent.Shipping` project, configure the publisher for the `OrderSubmittedEvent`. To do this use the `routing` object obtained when configuring the transport and add the following statement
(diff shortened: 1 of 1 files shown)


exercises/02-publish-subscribe/instructions.md
```diff
@@ -141,7 +141,7 @@ routing.RegisterPublisher(typeof(OrderSubmittedEvent), "Divergent.Sales");
 
 ### Step 8
 
-When Finance receives the `OrderSubmittedEvent` message it needs to keep track of item prices that belong to the submitted order. And finally initiate the payment process by sending the `InitiatePaymentProcessCommand` message.
+When Finance receives the `OrderSubmittedEvent` message it records prices for the items that belong to the submitted order. And finally initiate the payment process by sending the `InitiatePaymentProcessCommand` message.
```
tmasternak: Should we back-up the claim that we `need to keep track of item prices`?

dvdstelt: Good point. The idea is that items put in the shoppingcart for a certain price, remain in the cart for that price, even though the price might change during shopping.
This is a nice way to display how Sagas simplify this. Backing up the claim, for example/among others in the section `business requirements` will enhance this.

dvdstelt:
I think @tmasternak needs to add the text to "back-up the claim" mentioned in comments?

adamralph:
OK, I was a bit confused, because the claim we are talking about backing up is removed in this PR. However, I think get what you mean. Rather than remove that claim, we change it to a fact ("records prices for the items") and then we have a separate, specific sentence (or sentences) to explain why that happens?

tmasternak:
@adamralph done

exercises/02-publish-subscribe/instructions.md
```diff
@@ -141,7 +141,7 @@ routing.RegisterPublisher(typeof(OrderSubmittedEvent), "Divergent.Sales");
 
 ### Step 8
 
-When Finance receives the `OrderSubmittedEvent` message it needs to keep track of item prices that belong to the submitted order. And finally initiate the payment process by sending the `InitiatePaymentProcessCommand` message.
+When Finance receives the `OrderSubmittedEvent` message it records prices for the items that belong to the submitted order (this is needed to make sure that even when the product prices change, the client will be charged the amount first shown in the UI - value returned through `Divergent.Finance.API`). And finally initiate the payment process by sending the `InitiatePaymentProcessCommand` message.
```
dvdstelt: maybe 'initially shown in the UI'?

dvdstelt: I don't get this addition?

> value returned through `Divergent.Finance.API`

adamralph:
@dvdstelt @tmasternak I tweaked the changes. Please review.

PR #127, opened 2017-09-12
Remove "Yet" suffix from saga property name

SzymonPobiega:


SzymonPobiega:
Good catch @adamralph . I removed these two also.

PR #128, opened 2017-09-21
Fix compilation error in sample code

gieniowski:
Async keyword is not needed for methods that return Task.CompletedTask. In fact, this is compilation error.


mauroservienti:
Thanks @gieniowski, merged.

PR #130, opened 2017-10-05
Remove pagination support

mauroservienti:
Front end lists in workshop exercises are limited to 10 items. This means that if you test one of the exercises a few times, your latest changes are not shown, and it appears as though the solution isn't working.


mauroservienti:
@adamralph done for all now, also double checked that there are no references to pagination code in instructions documents.
Squash and merge?

mauroservienti:
squashed, please go ahead and merge.

PR #131, opened 2017-10-05
add build script

adamralph:


mauroservienti:
Can we in some way do the same thing that doco does also to configure the startup projects?

adamralph:
> Can we in some way do the same thing that doco does also to configure the startup projects?

AFAIK that works by producing .suo files which are bundled as part of the solution downloads. Given that the attendees clone the repo in order to attempt the exercises, I don't see at what point we could generate those .suo files.

adamralph:
@Particular/microservices-workshop-maintainers FYI I've switched on the setting to make a successful CI build mandatory for merging PR's to master.

mauroservienti:
> AFAIK that works by producing .suo files which are bundled as part of the solution downloads. Given that the attendees clone the repo in order to attempt the exercises, I don't see at what point we could generate those .suo files.

It seems that we can provide users with a script that will be downloaded when the repo is cloned and can be run locally. I'll investigate.

adamralph:
👍 sounds good!

PR #133, opened 2017-10-06
Packages update 04-integration

seanfarmar:
Updated packages:
NServiceBus.*
Microsoft.*
Log4Net
Common.Logging
Common.Logging.Core
Autofac
Newtonsoft.Json
NHibernate
Iesi.Collections
Nancy

### DIFF
--- a/exercises/04-integration/after/Divergent.Customers.API/Divergent.Customers.API.csproj
+++ b/exercises/04-integration/after/Divergent.Customers.API/Divergent.Customers.API.csproj
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="..\packages\Microsoft.Net.Compilers.2.0.1\build\Microsoft.Net.Compilers.props" Condition="Exists('..\packages\Microsoft.Net.Compilers.2.0.1\build\Microsoft.Net.Compilers.props')" />
-  <Import Project="..\packages\Microsoft.CodeDom.Providers.DotNetCompilerPlatform.1.0.3\build\Microsoft.CodeDom.Providers.DotNetCompilerPlatform.props" Condition="Exists('..\packages\Microsoft.CodeDom.Providers.DotNetCompilerPlatform.1.0.3\build\Microsoft.CodeDom.Providers.DotNetCompilerPlatform.props')" />
+  <Import Project="..\packages\Microsoft.Net.Compilers.2.3.2\build\Microsoft.Net.Compilers.props" Condition="Exists('..\packages\Microsoft.Net.Compilers.2.3.2\build\Microsoft.Net.Compilers.props')" />
+  <Import Project="..\packages\Microsoft.CodeDom.Providers.DotNetCompilerPlatform.1.0.7\build\net45\Microsoft.CodeDom.Providers.DotNetCompilerPlatform.props" Condition="Exists('..\packages\Microsoft.CodeDom.Providers.DotNetCompilerPlatform.1.0.7\build\net45\Microsoft.CodeDom.Providers.DotNetCompilerPlatform.props')" />
   <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
   <PropertyGroup>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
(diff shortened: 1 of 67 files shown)


exercises/04-integration/after/Divergent.Customers.API/Web.config
```diff
@@ -77,11 +77,10 @@
...
-      <compiler language="vb;vbs;visualbasic;vbscript" extension=".vb" type="Microsoft.CodeDom.Providers.DotNetCompilerPlatform.VBCodeProvider, Microsoft.CodeDom.Providers.DotNetCompilerPlatform, Version=1.0.3.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35" warningLevel="4" compilerOptions="/langversion:14 /nowarn:41008 /define:_MYTYPE=\&quot;Web\&quot; /optionInfer+" />
+      <compiler language="c#;cs;csharp" extension=".cs" type="Microsoft.CodeDom.Providers.DotNetCompilerPlatform.CSharpCodeProvider, Microsoft.CodeDom.Providers.DotNetCompilerPlatform, Version=1.0.7.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35" warningLevel="4" compilerOptions="/langversion:default /nowarn:1659;1699;1701" />
+      <compiler language="vb;vbs;visualbasic;vbscript" extension=".vb" type="Microsoft.CodeDom.Providers.DotNetCompilerPlatform.VBCodeProvider, Microsoft.CodeDom.Providers.DotNetCompilerPlatform, Version=1.0.7.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35" warningLevel="4" compilerOptions="/langversion:default /nowarn:41008 /define:_MYTYPE=\&quot;Web\&quot; /optionInfer+" />
```
adamralph: Are the changes to this file (and the similar changes to other config files) required for the package update?

exercises/04-integration/after/Divergent.Customers/EndpointConfig.cs
```diff
@@ -1,23 +0,0 @@
...
-            // NServiceBus will move messages that fail repeatedly to a separate "error" queue. We recommend
-            // that you start with a shared error queue for all your endpoints for easy integration with ServiceControl.
-            endpointConfiguration.SendFailedMessagesTo("error");
-
-            // NServiceBus will store a copy of each successfully process message in a separate "audit" queue. We recommend
-            // that you start with a shared audit queue for all your endpoints for easy integration with ServiceControl.
-            endpointConfiguration.AuditProcessedMessagesTo("audit");
-        }
-    }
-}
```
adamralph: Was this a redundant file? I.e. not included in the project?

exercises/04-integration/after/Divergent.Customers.API/Divergent.Customers.API.csproj
```diff
@@ -211,7 +210,7 @@
         <WebProjectProperties>
           <UseIIS>True</UseIIS>
           <AutoAssignPort>True</AutoAssignPort>
-          <DevelopmentServerPort>20185</DevelopmentServerPort>
+          <DevelopmentServerPort>20186</DevelopmentServerPort>
```
adamralph: Why is this change required? Is it related to the package upgrades?

seanfarmar: No, not related to the package update, but it didn't run properly without this change...

exercises/04-integration/after/Divergent.Customers/EndpointConfig.cs
seanfarmar: Not sure, I'll check.

exercises/04-integration/after/Divergent.Customers.API/Divergent.Customers.API.csproj
seanfarmar: I'll do the before as well, and squash all to one commit once all is good

exercises/04-integration/after/Divergent.Customers.API/Web.config
seanfarmar: These were updated when the microsoft.* packages where updated

exercises/04-integration/after/Divergent.Customers/EndpointConfig.cs
seanfarmar: yes, it was added by the host package and I accidentally committed it

exercises/04-integration/after/Divergent.Customers.API/Web.config
adamralph: OK

exercises/04-integration/after/Divergent.Customers/EndpointConfig.cs
adamralph: OK, please keep this change as a separate commit.

exercises/04-integration/after/Divergent.Customers.API/Divergent.Customers.API.csproj
adamralph: Please leave this change as a separate commit.

adamralph: BTW - I think we need a code change as a result of this. IIRC the host and port of each web API is hardcoded in C#. @Particular/microservices-workshop-maintainers am I correct?

What I can't understand is why I have had no trouble running the solution. What problem did you encounter before changing these ports?

seanfarmar: I had port collisions... so I looked at all the api endpoints and made sure they are correct...

exercises/04-integration/after/Divergent.Customers/EndpointConfig.cs
seanfarmar: I introduced this file while I was doing the updates AFAIK, so I squashed it...

adamralph: This file has to be present in master right now, otherwise it wouldn't show as a removal in this PR.

exercises/04-integration/after/Divergent.Customers.API/Divergent.Customers.API.csproj
adamralph: I don't have this problem. Everything runs OK when I run the solution as-is. I don't think I saw any workshop attendees having this problem either. @Particular/microservices-workshop-maintainers does anyone know why @seanfarmar is having this problem?

Regardless, would it be best to remove it from this PR and deal with this separately? If we can keep this PR isolated to package updates only then we can get it merged in quicker.

exercises/04-integration/after/Divergent.Customers/EndpointConfig.cs
seanfarmar: Yes you are right! It is on master...

What do you think we should do?

seanfarmar:
Reverted the ports change commit, so now we only have the packages updates for `before` and `after` solutions

it also includes the deletion of the redundant `EndpointConfig.cs` file

exercises/04-integration/after/Divergent.Customers/EndpointConfig.cs
adamralph: As discussed, I'm OK to mix this minor change with the package update.

seanfarmar:
@adamralph I pushed a version that removed the endpointconfig.cs file deletion... But i forced pushed it, so your changes have been overwritten... can you review again and re-introduce them please?

seanfarmar:
@adamralph LGTM :-)

PR #135, opened 2017-10-11
[WIP] Set Startup projects LINQPad script

mauroservienti:
This is a very handy way to setup startup projects in one go for all the exercises. This spike requires LINQPad, but the script can be easily converted into an executable that attendees can run on their machine once they download or clone this repo.

The `01-composite-ui-before.StartupProjects.txt` si not needed, it is in this PR just to show how it can be used to customize startup projects if needed. The `SetStartupProjects` is clever enough to detect automatically which projects should be startup projects. It works very well on all exercises.

### DIFF
--- a/SetStartupProjects.linq
+++ b/SetStartupProjects.linq
@@ -6,8 +6,7 @@
 void Main()
 {
 	var currentDiretory = Path.GetDirectoryName(Util.CurrentQueryPath);
-	var exercises = Path.Combine(currentDiretory, "exercises");
-	SetStartupProjects(exercises);
+	SetStartupProjects(currentDiretory);
 }
 
 public static void SetStartupProjects(string codeDirectory)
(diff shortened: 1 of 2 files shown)


dvdstelt:
Nice! We should try to figure out how we can define WebAPI projects to not pop-up a browser window and add that to `SetStartupProjects`. But for now this surely works nice!

mauroservienti:
that setting is in the project user file, that is not touched by this script. The easiest option is to convert them all to self hosted WebAPI Console Apps as we do for the demo. BTW we should convert them all to .NET Core as well as soon as we migrate to NServiceBus on .NET Core

adamralph:
Alternatively, could we just commit the .sou in a special folder, and then include a cmd file or similar which copies them to the `.vs` folder?

mauroservienti:
@Particular/microservices-workshop-maintainers given that the script can automatically guess startup projects and it requires no maintenance if we change solution structure, what about merging as is? it'll for sure alleviate our pain when developing and testing stuff due to the large amount of `git clean -fxd` I tend to perform during delopment

dvdstelt:
> convert them all to self hosted WebAPI Console Apps

I would rather create 1 console app to host them all. It's not the browser popups I hate, it's all the popups! :-)

adamralph:
Superseded by #205.

Issue #137, opened 2017-10-11
Avoid hard-coding routes in <a> tags


reashore:
For project Divergent.Website, the file /Views/Orders/Index.cshtml has a hard-coded route in the following <a> tag:

<a href="/Orders/Details/@order.OrderNumber">Number: #@order.OrderNumber</a>

It is a better coding practice to use an Html.ActionLink().

The loop on that page should be replaced with the following:

    @foreach (dynamic order in Model.Orders)
    {
        long orderNumber = order.OrderNumber;
        string orderNumberLink = $"Number: #{order.OrderNumber}";

        <tr>
            <td>@Html.ActionLink(orderNumberLink, "Details", "Orders", new {id = orderNumber })</td>
            <td>@order.ShippingStatus</td>
        </tr>
    }

mauroservienti:
thanks for raising this @reashore, I've raised #139 to fix this. #139 fixes only the ASP.Net Core demo since the legacy MVC one is destined to be deleted by #140

PR #139, opened 2017-10-12
Use html tag helpers

mauroservienti:
Connects to https://github.com/Particular/Workshop.Microservices/issues/137
Fix https://github.com/Particular/Workshop.Microservices/issues/137

### DIFF
--- a/demos/asp-net-core/Divergent.Website/Views/Orders/Index.cshtml
+++ b/demos/asp-net-core/Divergent.Website/Views/Orders/Index.cshtml
@@ -12,7 +12,7 @@
         @foreach (dynamic order in Model.Orders)
         {
             <tr>
-                <td><a href="/Orders/Details/@order.OrderNumber">Number: #@order.OrderNumber</a></td>
+                <td><a asp-controller="Orders" asp-action="Details" asp-route-id=@order.OrderNumber>Number: #@order.OrderNumber</a></td>
                 <td>@order.ShippingStatus</td>
             </tr>
 


demos/asp-net-core/Divergent.Website/Views/Orders/Index.cshtml
```diff
@@ -11,8 +11,11 @@
 
         @foreach (dynamic order in Model.Orders)
         {
+            long orderNumber = order.OrderNumber;
+            string orderNumberLink = $"Number: #{order.OrderNumber}";
+
             <tr>
-                <td><a href="/Orders/Details/@order.OrderNumber">Number: #@order.OrderNumber</a></td>
+                <td><a asp-area="" asp-controller="Orders" asp-action="Details" asp-route-id=@orderNumber>@orderNumberLink</a>
```
adamralph: Do the two new variables add enough value vs the single line option?

```html
<td><a asp-controller="Orders" asp-action="Details" asp-route-id=@order.OrderNumber>Number: #@order.OrderNumber</a>
```

mauroservienti: I'm on the fence, I like to have them split out for clarity, it's more readable IMO, but that portion of the code is not the main point of the demo, so we can change it as you suggest. No problems.

adamralph: One thing I definitely prefer is having `"Number: ..."` inside the `<a>` tag, otherwise the content is pulled away from the markup that surrounds it, purely because this is a templated file.

Assuming you agree, that would remove the need for the `orderNumberLink` variable, leaving only `orderNumber`. If you feel you'd rather have `@orderNumber` in the markup than `@order.OrderNumber` then I'd be OK with that.

mauroservienti: Code updated to match:

```
<td><a asp-controller="Orders" asp-action="Details" asp-route-id=@order.OrderNumber>Number: #@order.OrderNumber</a>
```

PR #145, opened 2017-10-13
[SPIKE] Appenders configuration

### LINKED ISSUE #144: BaseAddress is hard-coded
The base address is defined in the configuration files of the two "API.Host" projects.

However, it is hard-coded in the following files:
1) Divergent.Sales.ViewModelComposition.OrderDetailsViewModelAppender
2) Divergent.Shipping.ViewModelComposition.OrderDetailsViewModelAppender
3) Divergent.Sales.ViewModelComposition.OrdersListViewModelAppender

For example:

    string url = $"http://localhost:20296/api/shippinginfo/order/{id}";
	
So if the base address is changed then the application will fail.

mauroservienti:
Connects to #144 
Fix #144 

This PR allows Appenders to store configuration settings in an external configuration file, using the basic `IConfiguration` support provided by .NET Core.


adamralph:
Is it still worth doing this? Note that #144 was raised before we introduced [these comments](https://github.com/Particular/Workshop.Microservices/blob/d3d455ca04c725402f924cd8d306897a51dccf15/demos/asp-net-core/Divergent.Shipping.ViewModelComposition/OrderDetailsViewModelAppender.cs#L24).

mauroservienti:
on the fence, this entirely solves the problem and IMO is worth doing, but as said multiple times is a sample and thus out scope

dvdstelt:
Don't do it, I couldn't care less. I'm also not entirely sure what others are doing, but I've done the workshop and demos a few times now and saw Udi do it, but I've never ever heard anyone complain about the url being in there. I vote for closing this.

mauroservienti:
I'm fine with closing this, we can always reevaluate later, it's an easy change.

PR #147, opened 2017-10-17
Makes composition use OrderNumber instead of Id as a correlation key

mauroservienti:


demos/asp-net-core/Divergent.Sales.API.Host/Controllers/OrdersController.cs
```diff
@@ -9,18 +9,17 @@ namespace Divergent.Sales.API.Host.Controllers
     [RoutePrefix("api/orders")]
     public class OrdersController : ApiController
     {
-        [HttpGet]
-        public dynamic Get(int id)
+        [HttpGet, Route("{orderNumber}")]
```
adamralph: Why do we currently do not have a `Route` attribute here?

demos/asp-net-core/Divergent.Sales.API.Host/Controllers/OrdersController.cs
```diff
@@ -41,8 +40,7 @@ public IEnumerable<dynamic> Get(int pageIndex, int pageSize)
                     .Take(pageSize)
                     .Select(o => new
                     {
-                        Number = o.Id,
-                        o.Id,
+                        OrderNumber = o.OrderNumber,
```
adamralph: Must/should/could we change

```C#
.OrderBy(o => o.Id) //required by SQLite EF
```

to use `OrderNumber`?

mauroservienti: fixed

demos/asp-net-core/Divergent.Sales.API.Host/Controllers/OrdersController.cs
mauroservienti: because it was not needed, given that `id` is the default input parameter name for actions as stated in the default API configuration:

```
config.Routes.MapHttpRoute(
    name: "DefaultApi",
    routeTemplate: "api/{controller}/{id}",
    defaults: new { id = RouteParameter.Optional }
);
```

In order to change from `id` to `orderNumber` to make it explicit we need to instruct the WebAPI routing engine on how to match this route that is basically new.

demos/asp-net-core/Divergent.Sales.API.Host/Controllers/OrdersController.cs
```diff
@@ -36,13 +35,12 @@ public IEnumerable<dynamic> Get(int pageIndex, int pageSize)
             {
                 var orders = db.Orders
                     .Include(o => o.Items)
-                    .OrderBy(o => o.Id) //required by SQLite EF
+                    .OrderBy(o => o.OrderNumber) //required by SQLite EF
```
adamralph: Is the comment still relevant?

mauroservienti: For the sake of the sample/demo sorting is not required, it’s the EF implementation that requires sorting. He comment is there simply to remind that, we can remove it without any issue.

mauroservienti:
Naaa, I'll squash them

mauroservienti:
squashed, but now another commit from another branch popped up, but not locally, on my machine the other commit is still on the other branch only?

adamralph:
@mauroservienti it's because you squashed and rebased on to the `do-not-launch-iis-on-F5` branch instead of `master`.

mauroservienti:
No, I didn't, and it's now fixed. It happened yesterday as well, I pushed a change and it took 15 minutes to appear, then a I saw a GitHub status notification on Twitter mentioning queuing issues. Anyway, it should be OK now.

mauroservienti:
What's interesting is that I squashed from the laptop, where I don't have the `do-not-launch-iis-on-F5` branch

adamralph:
Weird. Something very strange must have happened inside GitHub then, because I checked the network graph and I could clearly see the rebased branch, with a single commit, based on top of the `do-not-launch-iis-on-F5` instead of `master`. Anyway, all looks good now. ¯\_(ツ)_/¯

PR #155, opened 2017-10-20
rename exercise instructions.md to README.md

adamralph:
It seems sensible that the instructions are shown when navigating to the folder in GitHub, rather than having to click through into `instructions.md`.


adamralph:
BTW - I double checked and there are no links to these documents, nor are they mentioned by name anywhere.

PR #162, opened 2017-10-23
IT/Ops API changes

adamralph:


demos/asp-net-core/ITOps.ViewModelComposition/EventHandler.cs
```diff
@@ -4,8 +4,5 @@
 
 namespace ITOps.ViewModelComposition
 {
-    interface ISubscription
-    {
-        Task Invoke(dynamic viewModel, object @event, RouteData routeData, IQueryCollection query);
-    }
+    public delegate Task EventHandler(dynamic pageViewModel, dynamic @event, RouteData routeData, IQueryCollection query);
```
mauroservienti: Subscribers now receives a `dynamic` event, where before with zero effort for the subscriber and one single `internal` interface and an `internal` class in the engine is was strongly typed and much better for subscribers

demos/asp-net-core/ITOps.ViewModelComposition/ISubscribeToCompositionEvents.cs
```diff
@@ -1,10 +1,7 @@
-﻿using Microsoft.AspNetCore.Http;
-using Microsoft.AspNetCore.Routing;
-
-namespace ITOps.ViewModelComposition
+﻿namespace ITOps.ViewModelComposition
 {
     public interface ISubscribeToCompositionEvents : IRouteInterceptor
     {
-        void Subscribe(ISubscriptionStorage subscriptionStorage, RouteData routeData, IQueryCollection query);
+        void Subscribe(IPublishCompositionEvents publisher);
```
mauroservienti: the removal of `RouteData` and, especially `IQueryCollection`, forces subscribers to always subscribe and if not interested discard the event when it's published. Not a big deal, but we could skip that, one less subscription to invoke for nothing.

The reason why previously `IQueryCollection` is passed to subscribers is because they can decide to subscribe or not based on QS content, a thing that cannot be done at `match` time since match results can be cached and thus should not be affected by QS. So, IMO, `IQueryCollection` is needed, and thus for consistency `RouteData` can be kept as well.

adamralph: Understandable, but in the context of the demo, is this YAGNI? We don't demo the use of the parameters. We ignore them in all cases.

demos/asp-net-core/ITOps.ViewModelComposition/EventHandler.cs
adamralph: Fair enough, this was my one concern with this approach, but I wasn't sure how much it mattered, since the subscribers already have to be "dynamic-aware" wrt to the view model. However, if the event can be typed, then it's true that we should probably do that. I'll restore the type safety.

adamralph:
@mauroservienti I pushed a fixup commit to restore the event type safety for subscribers.

demos/asp-net-core/ITOps.ViewModelComposition/ISubscribeToCompositionEvents.cs
mauroservienti: Well, in some way you're right, but we "marked" the engine as near to be production ready, and this change make it less, it prevents a legitimate use and it might also raise time wasting questions. Anyway if I accept expression bodied methods everywhere there is no reason to not accept this.

PR #167, opened 2017-10-25
Switch workshop exercise 1 before and after Sales endpoint from SQL to in-memory storage

seanfarmar:
Remove SQL Storage and replace it with in-memory objects.


exercises/01-composite-ui/after/Divergent.Sales.Data/Repositories/IOrderRepository.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Divergent.Sales.Data.Repositories
+{
+    using Models;
+
+    public interface IOrderRepository
+    {
+        Task<List<Order>> Orders();
```
dvdstelt: Why is this Task<T> ?  There's nothing awaitable and can be List<Order>

exercises/01-composite-ui/after/Divergent.Sales.Data/Repositories/OrderRepository.cs
```diff
@@ -0,0 +1,14 @@
...
+using System.Threading.Tasks;
+using Divergent.Sales.Data.Models;
+
+namespace Divergent.Sales.Data.Repositories
+{
+    public class OrderRepository : IOrderRepository
+    {
+        public async Task<List<Order>> Orders()
+        {
+            return OrderSeedData.Orders();
```
dvdstelt: We abstracted away OrderSeedData in a repository that's completely empty. Can we simplify this and move it to the repository?

seanfarmar: It would just make it easy to refactor later?

exercises/01-composite-ui/after/Divergent.Sales.Data/Repositories/IOrderRepository.cs
seanfarmar: I followed the pattern in the other projects... shall we change all of them maybe?

adamralph: Better to be aligned with the other projects in this PR. We can change all of them in another PR if we decide to.

exercises/01-composite-ui/after/Divergent.Sales.Data/Repositories/OrderRepository.cs
adamralph: It should really be done as part of this PR, to match the other projects.

exercises/01-composite-ui/before/Divergent.Sales.API/Controllers/OrdersController.cs
```diff
@@ -1,35 +1,32 @@
...
         }
 
         [HttpGet]
         public IEnumerable<dynamic> Get()
         {
-            var orders = _context.Orders
-                .Include(i => i.Items)
-                .Include(i => i.Items.Select(x => x.Product))
-                .ToArray();
+            var orders = _repository.Orders().Result;
```
adamralph: This method should be simply

```C#
public Task<List<Order>> Get() => _repository.Orders();
```

See `CustomersController` for reference. Note that we should not be mapping the `Order` object into a new anonymous type any longer. It should be a simple . This may result in some changes to the view model.

seanfarmar:
@dvdstelt @adamralph done as per comments

exercises/01-composite-ui/before/Divergent.Sales.Data/Repositories/OrderRepository.cs
```diff
@@ -0,0 +1,14 @@
...
+using System.Threading.Tasks;
+using Divergent.Sales.Data.Models;
+
+namespace Divergent.Sales.Data.Repositories
+{
+    public class OrderRepository : IOrderRepository
+    {
+        public async Task<List<Order>> Orders()
+        {
+            return OrderSeedData.Orders();
```
adamralph: This needs to be changed to match the `after` solution.

adamralph:
Also note that the build is failing with

> CSC : error CS2001: Source file 'C:\BuildAgent\work\8fd8bdad17a8f15e\exercises\01-composite-ui\after\Divergent.Sales.Data\Repositories\OrderSeedData.cs' could not be found. [C:\BuildAgent\work\8fd8bdad17a8f15e\exercises\01-composite-ui\after\Divergent.Sales.Data\Divergent.Sales.Data.csproj]

exercises/01-composite-ui/before/Divergent.Sales.API/Controllers/OrdersController.cs
adamralph: As discussed, I missed `ItemsCount = o.Items.Count` so we need to keep the mapping. The only required change is to change the method to be `async`.

seanfarmar:
Squashed :-)

PR #168, opened 2017-10-27
[WIP] Module 02 demos

dvdstelt:
Several demos that can be used in module 02.


mauroservienti:
@adamralph mode on:

> what about splitting this in multiple PRs, one per demo? And also giving meaningful names to projects? e.g. `ConsoleAppplication1` is not meaningful 😄 

@adamralph mode off.

mauroservienti:
oh, BTW, the build is failing

adamralph:
The build is failing because the solution names have spaces in them. I would prefer that we remove the spaces. While we could make the build script more resilient, I actually like the fact it's failing because of this. 😄

adamralph:
I would prefer that we use the `kebab-case` for the solution names, as we do for the exercise solutions.

dvdstelt:
build succeeded

dvdstelt:
This is still waiting to be approved? Can we do something about it, together with the DevEd PR that explain the demos?
I can even show them to presenters and/or create a recording of it.

PR #170, opened 2017-10-28
Fix exercise 1 "after" solution to match readme instructions

oysteinkrog:
After completing exercise 1 I noticed a difference between the before/after solution.
The exercise readme says to create ordersLoadedSubscriber.js, and provides the code snippet to put in this file.
In the after solution, this file is slightly different from the code snippet provided.


oysteinkrog:
Hmm, oops, I just noticed that this is the wrong file....

PR #173, opened 2017-11-06
[WIP] Switch exercise projects to self-hosting NServiceBus

### LINKED ISSUE #115: Move off NServiceBus.Host
given it will shortly be deprecated

seanfarmar:
Fixes #115


seanfarmar:
@mauroservienti @adamralph ping

seanfarmar:
Second version - converted the project to self-hosting.

@Particular/microservices-workshop-maintainers - Please review and if it's ok I'll do all the rest

exercises/04-integration/after/04-integration-after.sln
```diff
@@ -1,10 +1,7 @@
-﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 14
-VisualStudioVersion = 14.0.25420.1
+# Visual Studio 15
+VisualStudioVersion = 15.0.26730.16
 MinimumVisualStudioVersion = 10.0.40219.1
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Divergent.Customers", "Divergent.Customers\Divergent.Customers.csproj", "{02EE3114-734F-42DF-994C-98480E1489FE}"
-EndProject
```
adamralph: It looks like this file has become polluted due to the addition of `Customers.Host` in the first commit, and then the removal of the previous `Customers` and rename of `Customers.Host` to `Customers` in the second commit.

I think it's better if we treat this branch as a spike, to get feedback, but then the actual work is started again, on a fresh branch and in a new PR.

exercises/04-integration/after/Divergent.Customers/Config/EndpointConfig.cs
```diff
@@ -48,8 +32,12 @@ public void Customize(EndpointConfiguration endpointConfiguration)
...
 
             var conventions = endpointConfiguration.Conventions();
-            conventions.DefiningCommandsAs(t => t.Namespace != null && t.Namespace.StartsWith("Divergent") && t.Namespace.EndsWith("Commands") && t.Name.EndsWith("Command"));
-            conventions.DefiningEventsAs(t => t.Namespace != null && t.Namespace.StartsWith("Divergent") && t.Namespace.EndsWith("Events") && t.Name.EndsWith("Event"));
+            conventions.DefiningCommandsAs(t =>
+                t.Namespace != null && t.Namespace.StartsWith("Divergent") && t.Namespace.EndsWith("Commands") &&
+                t.Name.EndsWith("Command"));
+            conventions.DefiningEventsAs(t =>
+                t.Namespace != null && t.Namespace.StartsWith("Divergent") && t.Namespace.EndsWith("Events") &&
+                t.Name.EndsWith("Event"));
```
adamralph: It took me a while to figure out that there are no actual changes here and it's just the addition of line breaks. This is a good change, but please make it in a separate refactoring commit if it is going to be in the new PR.

exercises/04-integration/after/Divergent.Customers/ProgramService.cs
```diff
@@ -0,0 +1,100 @@
...
+
+        static ProgramService()
+        {
+            //TODO: optionally choose a custom logging library
+            // https://docs.particular.net/nservicebus/logging/#custom-logging
+            // LogManager.Use<TheLoggingFactory>();
+            logger = LogManager.GetLogger<ProgramService>();
+        }
+
+        public static void Main()
```
adamralph: I'd prefer not to conflate `Program.Main` and the `ServiceBase`. Can we keep them separate?

exercises/04-integration/after/Divergent.Customers/ProgramService.cs
```diff
@@ -0,0 +1,100 @@
...
+using System.ServiceProcess;
+using System.Threading.Tasks;
+using NServiceBus;
+using NServiceBus.Logging;
+using Divergent.Customers.Config;
+
+namespace Divergent.Customers
+{
+    [DesignerCategory("Code")]
+    class ProgramService : ServiceBase
```
adamralph: `ProgramService` is a weird name. I'd prefer `WindowsService`.

exercises/04-integration/after/Divergent.Customers/ProgramService.cs
```diff
@@ -0,0 +1,100 @@
...
+using System.ComponentModel;
+using System.ServiceProcess;
+using System.Threading.Tasks;
+using NServiceBus;
+using NServiceBus.Logging;
+using Divergent.Customers.Config;
+
+namespace Divergent.Customers
+{
+    [DesignerCategory("Code")]
```
adamralph: What is this?

exercises/04-integration/after/Divergent.Customers/ProgramService.cs
```diff
@@ -0,0 +1,100 @@
+﻿using System;
+using System.ComponentModel;
+using System.ServiceProcess;
+using System.Threading.Tasks;
+using NServiceBus;
+using NServiceBus.Logging;
+using Divergent.Customers.Config;
```
adamralph: Needs sorting.

exercises/04-integration/after/Divergent.Customers/ProgramService.cs
```diff
@@ -0,0 +1,100 @@
...
+    {
+        IEndpointInstance endpoint;
+        static ILog logger;
+
+        static ProgramService()
+        {
+            //TODO: optionally choose a custom logging library
+            // https://docs.particular.net/nservicebus/logging/#custom-logging
+            // LogManager.Use<TheLoggingFactory>();
+            logger = LogManager.GetLogger<ProgramService>();
```
adamralph: This should be a field initialiser rather than in a static constructor.

exercises/04-integration/after/Divergent.Customers/ServiceHelperClassic.cs
```diff
@@ -0,0 +1,12 @@
...
+namespace Divergent.Customers
+{
+    static class ServiceHelper
+    {
+        public static bool IsService()
+        {
+            return !Environment.UserInteractive;
+        }
+    }
+}
```
adamralph: This is used in one place only. I'd prefer to get rid of it and have the caller call `Environment.UserInteractive` instead.

exercises/04-integration/after/Divergent.Customers/ProgramService.cs
```diff
@@ -0,0 +1,100 @@
...
+using NServiceBus.Logging;
+using Divergent.Customers.Config;
+
+namespace Divergent.Customers
+{
+    [DesignerCategory("Code")]
+    class ProgramService : ServiceBase
+    {
+        IEndpointInstance endpoint;
+        static ILog logger;
```
adamralph: Both these should be prefixed by `private`.

mauroservienti: I'm used to prefix with private as well, so I'm fine with the suggestion. But isn't it redundant? fields are private by default.

mauroservienti:
> In addition to the inline comments, it looks ProgramService is based on some template with a lot of redundant comments and (sometimes empty) private methods. That should all be cleaned up.

I suppose that the template comes from https://github.com/Particular/NServiceBus.Bootstrap.WindowsService

exercises/04-integration/after/Divergent.Customers/ProgramService.cs
adamralph: I prefer to always be explicit. Different types of members have different access defaults and I'm loathed to have to map back to the default in my mind while I'm reading the code.

mauroservienti: `class` should be `internal` then 😉

adamralph: Yep, good catch! 😄

exercises/04-integration/after/04-integration-after.sln
seanfarmar: I will reset the head and do a clean conversion 👍

exercises/04-integration/after/Divergent.Customers/ProgramService.cs
seanfarmar: It comes from the template, I can clean all this if we need to

exercises/04-integration/after/Divergent.Customers/ServiceHelperClassic.cs
seanfarmar: @bording did a [PR to clean this template](https://github.com/Particular/dotnetTemplates/pull/3)

bording:
@adamralph as @seanfarmar mentioned, most of the things you're mentioning here come from the `dotnet new` template here: https://github.com/Particular/dotnetTemplates/blob/develop/src/Templates/NServiceBusWindowsService/ProgramService.cs

I've already made some changes, but if you think there are other things that we should alter as part of the template we'll be providing to customers, please suggest changes there.

exercises/02-publish-subscribe/after/Divergent.Customers/Handlers/OrderSubmittedHandler.cs
```diff
@@ -1,9 +1,9 @@
 ﻿using System.Data.Entity;
 using System.Threading.Tasks;
-using Divergent.Customers.Data.Context;
-using Divergent.Sales.Messages.Events;
 using NServiceBus;
 using NServiceBus.Logging;
+using Divergent.Customers.Data.Context;
+using Divergent.Sales.Messages.Events;
```
adamralph: This is an unrelated change. It should really be in a separate commit or PR. Same in a few other places.

exercises/02-publish-subscribe/after/Divergent.Customers/Config/EndpointConfig.cs
```diff
@@ -1,32 +1,25 @@
...
 using NServiceBus.Logging;
 using NServiceBus.Persistence;
-using ILog = Common.Logging.ILog;
-using LogManager = Common.Logging.LogManager;
 
 namespace Divergent.Customers.Config
 {
-    [EndpointName("Divergent.Customers")]
-    public class EndpointConfig : IConfigureThisEndpoint, AsA_Server
+    public class EndpointConfig
```
adamralph: This class should be static (same in other projects).

dvdstelt:
I definitely think we should definitely keep supporting VS2015 for as long as we can. Not everyone has the opportunity to upgrade because of various reasons.

adamralph:
At BuildStuff we still had some people (although a small minority) using VS 2015. I believe we should probably keep supporting it for now. Not everyone has a personal laptop where they can install VS 2017, nor a work laptop where VS 2017 is permitted (some orgs are still very slow to adopt new IDE versions).

mauroservienti:
Agreed. What's the impact on .NET Core projects, if any?

> Visual Studio 2015 will only support .NET Core <= 1.1 (ie. project.json). If you wish to use .csproj based .NET Core projects you will need to use Visual Studio 2017

bording:
> Agreed. What's the impact on .NET Core projects, if any?

@mauroservienti You have to have VS 2017 for .NET Core projects. AFAIK the old VS 2015 tooling cannot be used for a .NET Core 2.0 project.

adamralph:
As [discussed](https://github.com/Particular/Workshop/pull/193#issuecomment-350239333), we'll hold off merging this until after the merge of https://github.com/Particular/Workshop/pull/193

adamralph:
Superseded by #210

PR #190, opened 2017-11-24
[WIP] Polymer spa implementation

HEskandari:
This is a sample based on Polymer components. Although this is in a different folder, there are only 3 new projects (Sales/Shipping.UI and Frontend.spa) so all the other projects could be referenced from the aspnetcore folder if desired.


mauroservienti:
Thanks @HEskandari. Can you list in a comment all the projects that are NOT required for this demo, or much better directly drop them?

What I'd like to do is:

* remove from this demo all what is not really needed
* move the back-end services projects into a shared folder, e.g. `demos\shared-back-ends`, so that it'll be easier to add more demos

My idea is to end up with a structure such as the following:

```
demos\
   shared-back-ends
   polymer
   asp-net-core
   whatever
```

HEskandari:
Sure, I'll remove those projects and connect them to the version in aspnetcore directory. You should be able to easily move the shared project out.

HEskandari:
@mauroservienti Updated. Let me know if there's anything else.

mauroservienti:
Thanks @HEskandari, looks very good. @Particular/microservices-workshop-maintainers this is now stripped down to the minimum and is ready for review. However before proceeding I'd like to:

* merge the work we're doing to migrate exercises to Server Side Composition
* change this and the `asp-net-core` so that the Composition gateway is the same shared with exercises
* move demos back-ends into a shared folder `demos\shared-back-ends` and adjust both this and `asp-net-core` demo

these changes should make the repo much simpler to manager by reducing a lot of duplication

adamralph:
👍 sounds good @mauroservienti.

mauroservienti:
@HEskandari to simplify the review can you please squash the 2 commits so that everything boils down to polymer only?

HEskandari:
@mauroservienti done

HEskandari:
@mauroservienti arguable, the lib folder could also be removed, if we add instructions to do:

 - NPM Install first and foremost
 - Bower install after that

adamralph:
It sounds like we should indeed ignore the lib folders, given that they are the equivalent of a NuGet packages folder, and we don't commit that. I see that several projects have a lib folder and that they make up the bulk of the PR, so I assume it would shrink vastly in size if we ignored the lib folders?

mauroservienti:
there is a lot of debate in the npm/bower community about committing dependencies. It's mostly related to 2 things:

* SevVer is hard
* SevVer is harder in JavaScript

IMO if we remove any [range based dependency](https://github.com/npm/node-semver#advanced-range-syntax) from our bower and npm dependencies then we can safely ignore the lib folder, otherwise the risk is that different builds behave in different ways.

adamralph:
I'm 👎 to range based dependencies for this very reason.

mauroservienti:
@HEskandari can you please change dependencies to use a fixed range and then ignore and remove the lib folder as well.

PR #191, opened 2017-11-26
removed redundent link to script

oRosh1:


exercises/01-composite-ui/before/Divergent.Frontend/index.html
```diff
@@ -70,7 +70,6 @@
 
     <!-- Finance module -->
     <script src="/app/modules/finance/_module.js" type="text/javascript"></script>
-    <script src="/app/modules/finance/ordersLoadedSubscriber.js" type="text/javascript"></script>
```
dvdstelt: Why remove this line? Are you aware the exercise tries to add that file later on [over here](https://github.com/Particular/Workshop.Microservices/tree/master/exercises/01-composite-ui#step-2-1)?

dvdstelt: We could theoretically remove it, but then we'd need to update the exercise to add this line, otherwise the created javascript won't be loaded.
You can test in the "after" folder for this exercise. Remove it and the total for the order will never be retrieved.

oRosh1: Sorry I wasn't aware its part of the exercise.
Usually I check for errors to see if the build passed successfully.
Thank you for your serious comment.

seanfarmar: We will make the changes tomorrow

mauroservienti:
In light of the fact that we’re moving away from client side composition to API Gateway composition, I vote to close this one.

adamralph:
@oRosh1 thanks for raising this. As @mauroservienti mentions above, we are currently in the process of switching the exercises away from client composition using Angular to using server side composition, so we don't want to spend any time fixing issues relating to the current client side composition. For that reason, I'm closing this PR. Thanks again.

PR #192, opened 2017-11-27
SqlServer LocalDb on Windows 7

dvdstelt:
Users reported issues with LocalDb on Windows 7. I think the main reason is because SQL2016 doesn't support Windows 7.
You should therefor download & install SQL Express 2014 which contains LocalDb as well.

I should verify this before merging this, but maybe you know this is true and then you can just merge this ;-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,9 +179,9 @@ Using an elevated PowerShell prompt, run `Teardown-Databases.ps1` followed by `S
 
 Connect to the instance and run `Teardown-Databases.sql` followed by `Setup-Databases.sql`, both located in the [exercises/scripts](exercises/scripts) folder.
 
-### How can I use SQL Server LocalDb on Windows 7
+### Can I use Windows 7?
 
-Windows 7 isn't a supported operating system for SQL Server 2016. Instead install SQL Server 2014 Express, which you can (download here)[https://www.microsoft.com/en-us/download/details.aspx?id=42299].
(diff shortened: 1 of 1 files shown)


mauroservienti:
We don’t really have any constraint on the LocalDB version, thus whichever works on the user machine should work for us, I guess.

adamralph:
Marking as WIP due to:

> I should verify this before merging this, but maybe you know this is true and then you can just merge this ;-)

README.md
```diff
@@ -179,6 +179,10 @@ Using an elevated PowerShell prompt, run `Teardown-Databases.ps1` followed by `S
 
 Connect to the instance and run `Teardown-Databases.sql` followed by `Setup-Databases.sql`, both located in the [exercises/scripts](exercises/scripts) folder.
 
+### How can I use SQL Server LocalDb on Windows 7
+
+Windows 7 isn't a supported operating system for SQL Server 2016. Instead install SQL Server 2014 Express, which you can (download here)[https://www.microsoft.com/en-us/download/details.aspx?id=42299].
```
adamralph: > Windows 7 isn't a supported operating system for SQL Server 2016

The connection of this sentence to:

> How can I use SQL Server LocalDb on Windows 7

is not obvious.

Also, the text does not answer the question. It says you should use SQL Server 2014 Express instead. The title should probably be more general, e.g.

> Using Windows 7

adamralph:
@Particular/workshop-maintainers @dvdstelt I added a commit which generalises the FAQ. Please review.

PR #206, opened 2017-12-11
remove references to using a SQL Server instance

adamralph:
There's nothing stopping those attendees with SQL Server experience from doing this, but let's not encourage it. LocalDB is the simplest solution and should work for everyone.


adamralph:
@Particular/workshop-presenters any objections?

PR #210, opened 2017-12-29
replace NSB.Host in exercises

### LINKED ISSUE #115: Move off NServiceBus.Host

adamralph:
fixes #115


adamralph:
@seanfarmar @Particular/workshop-maintainers this is an alternative to #173

Marked as WIP, since I've only done the work for ex 2 after for initial review before investing any more effort.

exercises/02-publish-subscribe/after/Divergent.Customers/Program.cs
```diff
@@ -0,0 +1,37 @@
...
+namespace Divergent.Customers
+{
+    static class Program
+    {
+        public async static Task Main(string[] args)
+        {
+            var host = new Host(ConfigurationManager.ConnectionStrings[Host.EndpointName].ToString());
+
+            // pass this command line option to run as a windows service
+            if (args.Contains("--run-as-service"))
```
dvdstelt: Question : Why is this implementation so different from [our documentation](https://docs.particular.net/nservicebus/hosting/windows-service)?

exercises/02-publish-subscribe/after/Divergent.Customers/Program.cs
```diff
@@ -0,0 +1,37 @@
...
+            {
+                using (var windowsService = new WindowsService(host))
+                {
+                    ServiceBase.Run(windowsService);
+                    return;
+                }
+            }
+
+            Console.Title = Host.EndpointName;
```
dvdstelt: afaik this crashes when you're running as a service, but I might be mistaken.

exercises/02-publish-subscribe/after/Divergent.Finance/Host.cs
```diff
@@ -0,0 +1,77 @@
...
+using System.Threading.Tasks;
+using Divergent.Finance.PaymentClient;
+using Divergent.Sales.Messages.Events;
+using ITOps.EndpointConfig;
+using NServiceBus;
+using NServiceBus.Logging;
+
+namespace Divergent.Finance
+{
+    class Host
```
dvdstelt: This is almost a duplication from the Customers host.

exercises/02-publish-subscribe/after/Divergent.Finance/Host.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        public static string EndpointName => "Divergent.Finance";
+
+        public Host(string connectionString) => this.connectionString = connectionString;
+
+        public async Task Start()
+        {
+            try
+            {
+                var endpointConfiguration = new EndpointConfiguration(EndpointName)
```
dvdstelt: Seperate components are created so responsibilities aren't mixed, but then this mixes endpointconfig with the host and breaks SRP again?

exercises/02-publish-subscribe/after/Divergent.Finance/Host.cs
```diff
@@ -0,0 +1,77 @@
...
+                    .Configure(
+                        connectionString,
+                        routing => routing.RegisterPublisher(typeof(OrderSubmittedEvent), "Divergent.Sales"));
+
+                endpointConfiguration.RegisterComponents(registration =>
```
dvdstelt: Seperate components are created so responsibilities aren't mixed, but then this mixes DI with the host and breaks SRP again?

dvdstelt:
Is there a reason you went for a such a different approach than from our own documentation?
If we continue with this (or an alternate solution that's more like from our doco, if that's what we want), can't we provide an `EndpointConfiguration` to the host? Have a setup method in `Programs.cs` or somewhere else that sets up the config and container, before passing them on to the host?

exercises/02-publish-subscribe/after/Divergent.Customers/Program.cs
adamralph: @dvdstelt I believe that documentation needs updating. This implementation is based on our [dotnet template](https://github.com/Particular/dotnetTemplates/tree/develop/src/Templates/NServiceBusWindowsService).

exercises/02-publish-subscribe/after/Divergent.Customers/Program.cs
adamralph: When the endpoint is running as a service, the `--run-as-service` option should be passed, in which case this line will never be invoked.

exercises/02-publish-subscribe/after/Divergent.Finance/Host.cs
adamralph: It is, and I considered putting most of `Host.cs`, `WindowsService.cs` and `Program.cs` into shared files, but it started to feel like "over-re-use". In any case, this is no worse than the duplication we currently have in master with `ContainerSetup.cs` and `EndpointConfig.cs`. In fact, I'd argue that there's less duplication in this branch. I'd prefer we could merge as-is and address the duplication (which is effectively pre-existing) in a separate PR, should we choose to.

exercises/02-publish-subscribe/after/Divergent.Finance/Host.cs
adamralph: "Single responsibility" is usually a matter of opinion. 😉 

I'd argue that we know have a top level entry point (Program.cs), a windows service adapter (WindowService.cs) and a host for the endpoint (Host.cs). But in any case, this is following the pattern of our templates, so I'd prefer if we could merge as-is (to solve the pressing issue of using NSB.Host) and we address these concerns in the templates. If the templates are changed then we can carry those changes over into the exercises.

exercises/02-publish-subscribe/after/Divergent.Finance/Host.cs
adamralph: Same response as [previous](https://github.com/Particular/Workshop/pull/210/files#r159937815).

dvdstelt:
@adamralph Let's continue with this and complete it for all exercises.

Issue #214, opened 2018-01-16
Use SqlPersistence instead of NHibernate persistence


jbogard:
Should simplify the saga data as well and remove any questions of "do we have to use NHibernate"

jbogard:
Maybe I didn't see it - what's the problem with using `SqlSaga`?

adamralph:
@jbogard `SqlSaga` is a special case just for the SQL persistence. At the moment we're considering how we can remove this wrinkle so that we can return to a single base class regardless of persistence choice. It's not a huge problem, but personally I'm a bit hesitant to showcase it in the workshop, given that it's ideally something we'd like to get rid of.

DavidBoike:
Completed in #279

PR #216, opened 2018-01-16
[WIP] switch to SQL persistence

adamralph:
Relates to #214.

### DIFF
--- a/exercises/03-sagas/after/Divergent.Shipping/Divergent.Shipping.csproj
+++ b/exercises/03-sagas/after/Divergent.Shipping/Divergent.Shipping.csproj
@@ -43,14 +43,21 @@
       <HintPath>..\packages\EntityFramework.6.1.3\lib\net45\EntityFramework.SqlServer.dll</HintPath>
       <Private>True</Private>
     </Reference>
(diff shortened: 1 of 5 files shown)


adamralph:
I've only done the ex 3 after solution for now, as a POC. I've tested the solution and it runs just fine.

The only thing I'm hesitant about is the need to switch to `SqlSaga`, as this introduces another complication in the context of the workshop.

mauroservienti:
> The only thing I'm hesitant about is the need to switch to `SqlSaga`, as this introduces another complication in the context of the workshop.

Same reason why I’m hesitant too.

dvdstelt:
You should really start with ex4 after, instead of ex3. Then you cover every part of the entire workshop.

adamralph:
@dvdstelt this is really just a spike, to see how the conversion would look.

DavidBoike:
@adamralph You did it right, but I agree that I'd be hesitant to push it in workshop with the current required side-effects.

adamralph:
This spike has served its purpose. For now, we won't switch to SQL persistence. We can revisit this if/when we remove `SqlSaga`.

PR #221, opened 2018-01-29
remove Common.Logging.Log4Net1211

adamralph:


exercises/shared-api-gateway/ITOps.EndpointConfig/EndpointConfigurationExtensions.cs
```diff
@@ -18,7 +16,7 @@ public static class EndpointConfigurationExtensions
             string connectionString,
             Action<RoutingSettings<MsmqTransport>> configureRouting)
         {
-            NServiceBus.Logging.LogManager.Use<DefaultFactory>();
+            LogManager.Use<DefaultFactory>();
```
DavidBoike: Is this even necessary?

adamralph: I've no idea.

@Particular/logging-maintainers any idea what this is for? Completely redundant?

I guess I'll just try removing it and if I still get logs, all good!

mauroservienti:
What’s the purpose of the `EndpointConfigurationExtensions` class? And why is it in the `shared-api-gateway` folder, that contains only API Gateway related code?

adamralph:
@mauroservienti `EndpointConfigurationExtensions` applies all the endpoint config that common to all endpoints.

`shared-api-gateway/` is indeed mis-named, since the `ITOps.EndpointConfig` project was added to it. I'll rename it to `shared` in another PR. 😉

mauroservienti:
Well, it’s the other way round since it was meant to contain API Gateway code only,  it I’m fine with making it more general purpose by renaming it to `shared` only

exercises/shared-api-gateway/ITOps.EndpointConfig/EndpointConfigurationExtensions.cs
adamralph: Removed, and logs still appeared.

adamralph:
Commit amended and WIP removed.

PR #223, opened 2018-01-29
refactor API controllers

adamralph:
- switched usage of `ToArray()` to `ToList()` (less allocation)
- renamed variables with names starting with an underscore
- renamed variables with single char names
- simplified `Split()` call
- minimised EF context scope
- aligned EF context variable naming
- in-lined single use local variables
- converted one line constructors to expressions
- changed list property assignment to eager evaluation


adamralph:
Marked as WIP since I've only done the `after` solution so far. I'd like to get a review on these refactorings before applying them to the `before` solutions.

adamralph:
I'll go ahead and do the `before` solutions.

adamralph:
@DavidBoike I amended the commit to include the changes to the `before` solutions. I also did a directory diff between each `before` and `after` solution on this branch and confirmed that the diffs are the same as those on `master`.

adamralph:
Also note the edit to the ex 01 README.md.

PR #225, opened 2018-01-31
remove redundant ToArray calls

adamralph:


mauroservienti:
Have you tested it? on my machine (when we were still using .NET Core 1.x) without the `ToArray()` is was causing painful issues to the razor view engine when trying to iterate over a dynamic enumerable not yet evaluated when rendering the page.

adamralph:
@mauroservienti I ran 04 after and the website demo, and it seemed to work fine.

I just tried to run the demos, but the website doesn't seem to run (only the API hosts) and the composition gateway exits immediately after starting. 😕 

Marking as WIP until the demos have been tested.

adamralph:
@Particular/workshop-maintainers I smoke tested the demos (both website and gateway) and everything worked fine. This should be good to go now.

mauroservienti:
tested locally, since on my machine it was failing a few months ago. All fine.

PR #236, opened 2018-03-19
[WIP] move providers to *.ITOps projects

adamralph:


mauroservienti:
ITOps takes care of the hosting concern, from the point of view of Customers and Shipping it’s not ITOps, it’s about the data they provide. The `IProvide*` interace implementation should not probably be in the `*.Data` assembly but I dont’ think the assembly should be named `*.ITOps`.
I’d rather name them `*.Data.Providers` if we really want to split them out.

adamralph:
@mauroservienti the two projects already know about ITOps, since they reference the assembly which defines the interface. They are service-specific implementations of things defined in ITOps, so the name, `Divergent.{Service}.ITOps` feels natural to me.

I don't think they should have `Data` in their names since "data" is private to a service.

exercises/04-integration/after/Divergent.Customers.ITOps/Customer.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace Divergent.Customers.ITOps
+{
+    internal class Customer
```
danielmarbach: In this case the EF Context used here could directly map to `CustomerInfo` and then we wouldn't even need this type.

exercises/04-integration/after/Divergent.Customers.ITOps/CustomerInfoProvider.cs
```diff
@@ -2,13 +2,13 @@
...
+namespace Divergent.Customers.ITOps
 {
     public class CustomerInfoProvider : IProvideCustomerInfo
     {
         public Task<CustomerInfo> GetCustomerInfo(int customerId)
         {
-            using (var db = new Context.CustomersContext())
+            using (var db = new CustomersContext())
             {
                 var customer = db.Customers.Where(c => c.Id == customerId).Single();
```
danielmarbach: Minor: This could be async by using `SingleAsync`

mauroservienti:
> the two projects already know about ITOps, since they reference the assembly which defines the interface. They are service-specific implementations of things defined in ITOps, so the name, Divergent.{Service}.ITOps feels natural to me.

I disagree @adamralph, using the same reasoning we should name projects that contain `ViewModel Appenders` to have the `ITOps` suffix somewhere simply because they are hosted in something named `ITOps Composition Gateway` and because they reference something that comes from ITOps?

They are not ITOps related, they do not serve any ITOps purpose, they are simply hosted at runtime by a service owned by ITOps. Their main and only purpose is to provide data to compose the FedEx XML document.

dvdstelt:
Why is `CustomersContext` duplicated in the `Divergent.Customers.ITOps` project? This is code duplication and it also doesn't add the data seeding and foreign key definition for the table?

The .csproj is also of a different format than the older projects. I could not care less, but these things tend to be a major issue to some of us. So I wonder why this was done.

That's what I found at first sight, but I will review it more properly later.

adamralph:
@mauroservienti OK, fair point. How about `Divergent.{Service}.Providers`?

mauroservienti:
I'm good with that, yes.

adamralph:
@dvdstelt 

> Why is CustomersContext duplicated in the Divergent.Customers.ITOps project? This is code duplication and it also doesn't add the data seeding and foreign key definition for the table?

That is deliberate. The new project only does the minimum required to transform what is currently in the database to the satisfy the provider interface.

> The .csproj is also of a different format than the older projects. I could not care less, but these things tend to be a major issue to some of us. So I wonder why this was done.

We already have VS2017 as a requirement so it's OK to use the new project format. I figured that a brand new project can use the new format. The older projects can be converted as a separate exercise. I don't think making the new project use the old format has any benefit. You could argue consistency, but I believe the cost of yet another old style project outweighs that.

exercises/04-integration/after/Divergent.Customers.ITOps/Customer.cs
adamralph: @danielmarbach that sounds good, but I know next to nothing about EF. How can that be done?

adamralph:
@dvdstelt I was mistaken. I thought, from your comment, that I must have used the new project format, but in fact, it's still the old one, but it uses `PackageReference` instead of `packages.config`. Still, the same argument applies.

dvdstelt:
> The new project only does the minimum required to transform what is currently in the database to the satisfy the provider interface.

I can't agree with that.

- Both fall within the responsibility of the same service boundary.
- Both fall within the same team responsible for the code.
- Both don't get exposed outside.
- The only difference is that it provides a single less property. As a result a fully build system should contain dozens and dozens of contexts to satisfy every scenario. We won't do that either.
- In my opinion, it's simple code duplication and over-engineering.
- It's also not complete because it still does not satisfy the seeding code and could thus break.

danielmarbach:
@dvdstelt @mauroservienti @adamralph a quick call to hash it out?

dvdstelt:
Sounds good. Adam and Mauro are not online until this afternoon. When Adam comes online, I have two meetings, of which one is with a customer.
I'll create a short meeting for 5pm and we'll see if everyone can join.

danielmarbach:
I can only do tomorrow

adamralph:
Closing for now since we've failed to come to consensus, and it seems to be a low priority concern.

PR #237, opened 2018-03-21
use event constructors

adamralph:


exercises/02-publish-subscribe/README.md
```diff
@@ -73,14 +73,14 @@ namespace Divergent.Sales.Messages.Events
...
+Have a look at `SubmitOrderHandler` class in the `Divergent.Sales` project. At the end of the `Handle` method, publish the `OrderSubmittedEvent`, by calling the ```context.Publish()``` method. Copy the properties from the incoming `SubmitOrderCommand` message, to the properties of the event.
 
 ```c#
-await context.Publish<OrderSubmittedEvent>(e =>
+await context.Publish(new OrderSubmittedEvent
 {
-    e.OrderId = order.Id;
-    e.CustomerId = message.CustomerId;
-    e.Products = message.Products;
+    OrderId = order.Id;
```
DavidBoike: You still have semicolons in here

adamralph: Oops! Good catch. Since the change was so trivial, I amended the commit directly. Should be good now.

mauroservienti:
Minor nitpick. Isn't the last comma when using object initializers redundant? Instead of

```
new Foo()
{
   Property = 123,
}
```

It should be 

```
new Foo()
{
   Property = 123
}
```

Without the comma after last property assignament

adamralph:
@mauroservienti my preference is to always have a trailing comma, even on the last, or only line. That way, each line is self contained. I.e.: When I add a line, I don't have to edit the previous line. When I move lines, I don't have to play with commas to fix things up.

PR #239, opened 2018-04-10
Enable DB scripts working with SQL Local DB 2017 and simplify DB setup

danielmarbach:
- Switches to promote script to local db with SQLCMD
- Adds explanation if necessary to setup local db

PR #241, opened 2018-04-16
Fix database connection path

danielmarbach:
Fixes regression after #239


dvdstelt:
Question: What is the reason you changed it to 'particular-workshop' in the first place? Besides the fact that we have 2 workshops? Would it not also have been easier to modify the scripts to change it back, instead of every single config file?

danielmarbach:
@dvdstelt https://github.com/Particular/Workshop/pull/239#discussion_r180755035

@adamralph mentioned the intention was to get entirely rid of the word "microservices". I fixed the last mentioning and @adamralph merged the PR. Given that previous work has been done to get rid of the wording and the last PR did that in the database name but not in the web.config the logical consequence is to fix it in the web configs

PR #247, opened 2018-05-07
move demos to net core 2.0 runtime

mauroservienti:
Moves the demo to NET Core 2.x

Since it was an easy fix I took the liberty of working on it. We should probably merge this after NDC Minnesota to not create any confusion to attendees

### DIFF
--- a/demos/asp-net-core/README.md
+++ b/demos/asp-net-core/README.md
@@ -2,7 +2,7 @@
 
 This demo is composed of two solutions which demonstrate UI Composition techniques using ASP.NET Core.
 
-In order to run the demos, you require the latest .NET Core Runtime in the [1.1.x version range](https://www.microsoft.com/net/download/all).
+In order to run the demos, you require the latest .NET Core Runtime in the [2.0.x version range](https://www.microsoft.com/net/download/all).
 
 ### Divergent.CompositionGateway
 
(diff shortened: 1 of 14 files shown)


demos/asp-net-core/README.md
```diff
@@ -2,7 +2,7 @@
 
 This demo is composed of two solutions which demonstrate UI Composition techniques using ASP.NET Core.
 
-In order to run the demos, you require the latest .NET Core Runtime in the [1.1.x version range](https://www.microsoft.com/net/download/all).
+In order to run the demos, you require the latest .NET Core Runtime in the [2.x version range](https://www.microsoft.com/net/download/all).
```
adamralph: I'm not sure that `netcoreapp2.0` projects will run against the 2.1 runtime. This may have to be the 2.0.x version range.

mauroservienti: are you saying that runtime are not backward compatible? :trollface:

mauroservienti: fixed

mauroservienti:
> I believe the build is failing because something is going wrong with package restore. It may be worth upgrading to the latest NuGet.exe (4.6.2) or switching to dotnet restore.

what should I do to fix it?

mauroservienti:
rebased on `master`. @adamralph @DavidBoike @indualagarsamy please take a look

mauroservienti:
Removed the `PackageTargetFallback` setting, root cause of the build failure, from the `Divergent.Website` csproj file. I have no idea why it was there 😄

mauroservienti:
build is green now 😄

mauroservienti:
@adamralph I amended the commit to remove the `orig` file. I also dropped App Insights that is not really needed.

PR #249, opened 2018-05-07
[WIP] initial idea for a script to run demos

mauroservienti:
I'm sick of VS failing on me when running demos and exercises. This si a very raw idea to create scripts to run each demo and each exercise from command line.

I'm sure that the target can be greatly improved however my knowledge of the tolling we're using is near to zero.

### DIFF
--- a/build.csx
+++ b/build.csx
@@ -2,6 +2,7 @@
 
 #load "packages/simple-targets-csx.6.0.0/contentFiles/csx/any/simple-targets.csx"
 #load "scripts/cmd.csx"
+#load "scripts/start.csx"
 
 using System;
 using static SimpleTargets;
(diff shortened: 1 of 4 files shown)


build.csx
```diff
@@ -92,4 +92,38 @@ targets.Add(
     }
 );
 
+targets.Add(
+    "run-gateway-demo",
+    DependsOn("find-msbuild", "restore-demos"),
```
adamralph: This line should be:

```C#
DependsOn("demos"),
```

I.e. you want to build the demos before running the API host exe's.

build.csx
```diff
@@ -92,4 +92,38 @@ targets.Add(
...
+    () =>
+    {
+        var salesAPIHost = new Process();
+        salesAPIHost.StartInfo = new ProcessStartInfo
+        {
+            FileName = @".\demos\asp-net-core\Divergent.Sales.API.Host\bin\debug\Divergent.Sales.API.Host.exe",
+            UseShellExecute = true
+        };
+        
+        salesAPIHost.Start();
```
mauroservienti: @adamralph instead of this verbosity here should I just add a `CmdNoWait` command in the `cmd.csx` script?
Because if we wanna add a target here for every demo and every exercise, the verbosity will explode

build.csx
mauroservienti: thanks, I'll fix it

mauroservienti:
@adamralph Is there a way to pass arguments to a target?
Say that I wanna do: `build.cmd <target-name> <args-for-the-target>`

mauroservienti:
I've simplified the thing introducing a start script. Do you think it's also worth it to simplify paths by doing some magic?

Should I go ahead and add other scripts for all the exercises and the second demo?

mauroservienti:
😭 https://github.com/dotnet/cli/issues/8487 😞

mauroservienti:
Horrendous https://github.com/Particular/Workshop/pull/249/commits/09780ce033114e4c3739b3d4aeef75ddae59735e
 Cannot find any better solution, since `dotnet run` will take some time to startup the webserver the browser will open the page before the webserver is ready, thus 404. It's enough to wait for the webserver to be ready and then refresh the browser.

mauroservienti:
Another option is obviously to not automatically start the browser. The URL is displayed in the console window, so it's enough to copy/paste it.

adamralph:
Ugh, this is an ugly one. I've given a 👍 to https://github.com/dotnet/cli/issues/8487, FWIW.

adamralph:
@mauroservienti I'm not sure how useful it will be. We will still probably open VS to show the code, no? If it's a case of getting the weird debug error the first time, then I'll probably prep my machine with an F5 before delivering the module, so I know it will work when I hit F5 later. If it "just worked" then I might be tempted to use it, but the browser launch thing is awkward.

mauroservienti:
for me, as of now, it's a huge advantage. My laptop is not that powerful and if I can avoid using VS just to run something it's much better. Things might change in the future and a new laptop might be bought. But not being my primary machine I'm not interested in investing too much money.

Anyway we can leave this pending for now.

run-asp-net-gateway-demo.bat
```diff
@@ -0,0 +1 @@
+build.cmd run-gateway-demo
```
adamralph: minor nitpicks:

- The file extension should be `.cmd` to match the existing ones
- The filename and the target named aren't aligned as with `set-startup-projects`, perhaps rename the file to `run-gateway-demo.cmd`?
- This line should be terminated with a CRLF

run-asp-net-website-demo.bat
```diff
@@ -0,0 +1 @@
+build.cmd run-website-demo
```
adamralph: Same minor nitpicks as the other file.

mauroservienti:
Withdrawn.

PR #251, opened 2018-05-07
[WIP] Ignore .vs folders

mauroservienti:

PR #252, opened 2018-05-07
Minimal .gitignore

mauroservienti:
Should replace https://github.com/Particular/Workshop/pull/251


.gitignore
```diff
@@ -1,160 +1,19 @@
...
-*.vsp
-*.vspx
-
-# Guidance Automation Toolkit
-*.gpState
+# Nuget packages
+packages/
 
 # ReSharper is a .NET coding add-in
+
```
adamralph: Why has this line been added?

mauroservienti:
I marked it as wip so to decide if we wanna go with this or #251 , this is it.

.gitignore
mauroservienti: It happened to me that sometimes R# detritus are included. I’ve not been using R# for a long time now, but there are people in the company that do use it. Are you using it @adamralph ?

adamralph: Sorry, I didn't make myself clear. I just meant the extra blank line after the comment. The other sections don't have it.

No, I don't use R#. 😉

mauroservienti: A typo

mauroservienti: Fixed

dvdstelt:
Personal question: Why would we want a minimalistic .gitignore file? Is there any benefit?

Devil's Advocate: There should be an issue for this, but there is none. Not even a description?

adamralph:
The benefit is simplicity and removing YAGNI cruft.

I'm not sure we need an issue for such a trivial change.

dvdstelt:
> The benefit is simplicity

How could it be more simple than let it be like Visual Studio ot Git created it?

> removing YAGNI cruft

It's YAGNI to remove it!

mauroservienti:
The one created by default by GitHub or Visual Studio contains all possibile exclusions for all possible languages, including things like C++ or ClickOnce deployments. Whenever we need to manage it it takes more time to understand what’s in there then to edit it.

PR #255, opened 2018-05-08
set-startup-projects.cmd requires vs to be closed

mauroservienti:


README.md
```diff
@@ -83,6 +83,8 @@ The exercises are contained in eight Visual Studio solutions under [exercises](e
 
 Before opening any exercise solutions, set the startup projects by navigating to your copy of this repo and running `set-startup-projects.cmd`. Note that if you `git clean` your clone, you will have to run this command again.
 
+Note: do not run the `set-startup-projects.cmd` with any exercise or demo solution open. The script deletes and then regenerates solutions user setting, if run with a solution open it might corrupt solution user settings.
```
adamralph: The grammar needs some tweaking, perhaps:

> Note: do not run `set-startup-projects.cmd` while an exercise or demo solution is open. The script deletes and regenerates user settings for every solution. If it is run while a solution is open, the settings may be corrupted.

mauroservienti: Just change it and push Adam, it was late night when I edited it :smile:

PR #256, opened 2018-05-08
Fix Fedex Request null XDocument

DavidBoike:
CallFedexWebService method is a no-op so it never broke anything, but an attendee actually examined with the debugger and found that the XDocument was null.

### DIFF
--- a/exercises/04-integration/README.md
+++ b/exercises/04-integration/README.md
@@ -105,15 +105,16 @@ public class ShipWithFedexCommandHandler : IHandleMessages<ShipWithFedexCommand>
 
     private XDocument CreateFedexRequest(string name, string street, string city, string postalCode, string country)
     {
-        var shipment =
-            new XElement("FedExShipment",
-                new XElement("ShipTo",
-                    new XElement("Name", name),
-                    new XElement("Street", street),
-                    new XElement("City", city),
-                    new XElement("PostalCode", postalCode),
-                    new XElement("Country", country)));
-        return shipment.Document;
+        var shipment = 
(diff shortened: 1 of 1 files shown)


DavidBoike:
Need to update the after solution to match.

mauroservienti:
@DavidBoike it's probably only instructions, since the code in the after solution is already fine

https://github.com/Particular/Workshop/blob/d6a2ca5bb248908b5775a04e1d34e19ed95e3776/exercises/04-integration/after/Divergent.ITOps/Handlers/ShipWithFedexCommandHandler.cs#L36-L48

exercises/04-integration/README.md
```diff
@@ -105,14 +105,14 @@ public class ShipWithFedexCommandHandler : IHandleMessages<ShipWithFedexCommand>
...
-        var shipment =
+        var shipment = new XDocument(
             new XElement("FedExShipment",
                 new XElement("ShipTo",
                     new XElement("Name", name),
                     new XElement("Street", street),
                     new XElement("City", city),
                     new XElement("PostalCode", postalCode),
-                    new XElement("Country", country)));
+                    new XElement("Country", country))));
```
adamralph: It's an improvement, but it's still different to what's in the after solution (see next line). We should probably copypasta this from https://github.com/Particular/Workshop/blob/d6a2ca5bb248908b5775a04e1d34e19ed95e3776/exercises/04-integration/after/Divergent.ITOps/Handlers/ShipWithFedexCommandHandler.cs#L36-L48

PR #258, opened 2018-05-08
[WIP] Surface composition exceptions

mauroservienti:
When exceptions happen during the composition process they are returned to the client as `HTTP-500` with no other details than `Internal Server Error`.

This is an attempt to surface composition exceptions. It's currently very raw, it's merely an exercise, and serves only as a discussion. The problem we want to solve is: when attendees are doing exercises it's very easy to mess up something or in the composition engine or in the back-end APIs. In this case the UI is not helpful at all since it simply say that something went wrong, without any hint on what went badly.

By using something similar to what is proposed here, the result in the browser console will be:

image

In order to go ahead some async/await-fu is needed, I'm pretty sure there is a better solution to this than the one I applied.


mauroservienti:
Closing in favor of https://github.com/Particular/Workshop/pull/263

PR #260, opened 2018-05-09
make CustomerInfoProvider async

danielmarbach:


exercises/04-integration/after/Divergent.Customers.Data/ITOps/CustomerInfoProvider.cs
```diff
@@ -1,25 +1,26 @@
-﻿using System.Linq;
+﻿using System.Data.Entity;
```
adamralph: This is the one thing that could trip up the attendees. Without it, the IDE seems to have no idea where `SingleAsync` lives and does not provide a quick fix to add it. This confused the hell out of me last time it happened to me. Perhaps it needs to go into the instructions somewhere?

danielmarbach: @adamralph I added a note. What do you think?

danielmarbach:
@adamralph changed

PR #263, opened 2018-05-09
[WIP] Add logging to composition engine

mauroservienti:
Replaces https://github.com/Particular/Workshop/pull/258

it's a possible implementation to provide logging at the composition engine level, the same logic can be applied to all back-end APIs, so that all errors will surface somewhere

### DIFF
--- a/exercises/shared-api-gateway/ITOps.ViewModelComposition/CompositionHandler.cs
+++ b/exercises/shared-api-gateway/ITOps.ViewModelComposition/CompositionHandler.cs
@@ -1,6 +1,7 @@
 ﻿using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
@@ -11,7 +12,8 @@ namespace ITOps.ViewModelComposition
     {
         public static async Task<(dynamic ViewModel, int StatusCode)> HandleGetRequest(HttpContext context)
         {
-            var viewModel = new DynamicViewModel(context.GetRouteData(), context.Request.Query);
+            var loggerFactory = context.RequestServices.GetService<ILoggerFactory>();
+            var viewModel = new DynamicViewModel(context.GetRouteData(), context.Request.Query, loggerFactory);
             var routeData = context.GetRouteData();
 
             // matching interceptors could be cached by URL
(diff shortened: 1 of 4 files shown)


exercises/shared-api-gateway/ITOps.ViewModelComposition/CompositionHandler.cs
```diff
@@ -30,7 +32,17 @@ public static async Task<(dynamic ViewModel, int StatusCode)> HandleGetRequest(H
...
                 foreach (var appender in interceptors.OfType<IViewModelAppender>())
                 {
-                    pendingTasks.Add(appender.Append(viewModel, routeData, context.Request.Query));
+                    pendingTasks.Add(appender.Append(viewModel, routeData, context.Request.Query)
+                        .ContinueWith(t =>
+                        {
+                            t.Exception.Handle(ex =>
+                            {
+                                var logger = loggerFactory.CreateLogger(appender.GetType());
+                                //adjust to emit the correct information
```
adamralph: What does this comment mean?

mauroservienti: Because I’m not sure that Exception.ToString() is the right thing to do in this case. I need some @danielmarbach -fu 😄

mauroservienti: That was my first attempt and I like it much more. However at that point we’ve lost the connection between the failed Task and the failing Appender/Handler. We’re using the appender/handler type to create a specific logger

adamralph:
What if we attached something at the HTTP pipeline level which logged all exceptions (together with stack trace) to the console before sending out a 500. Would that also work?

danielmarbach:
I'll try to send in a proposal when I sit at the gate

Mauro Servienti <notifications@github.com> schrieb am Mi., 9. Mai 2018,
21:32:

mauroservienti:
@adamralph yes, that would also work, however:

* When the exception is already up in the http pipeline we've lost all the context of where it was raised, by attaching a logger to each task we can log details about the failing handler/appender
* it was similar to what I was doing with the previous attempt (#258), and it requires "hacks" like https://github.com/Particular/Workshop/blob/2d39caf3099b0c41f12f4ce3f563b82fda864041/exercises/shared-api-gateway/ITOps.ViewModelComposition/DynamicViewModel.cs#L78-L85 to make **all** the to bubble up

exercises/shared-api-gateway/ITOps.ViewModelComposition/CompositionHandler.cs
danielmarbach: @mauroservienti @adamralph what do you think about this https://github.com/Particular/Workshop/pull/264

adamralph:
If the console output included the stack trace, then we'd see which handler threw it.

The console output could also loop through all the exceptions contained within an `AggregateException` to remove the need for the flatten hack in `DynamicViewModel`.

mauroservienti:
As soon as something is awaited the aggregate exception will only contain 1 exception and all the others are gone, this is why there is the Flatten horribleness

adamralph:
Raised #268 as an alternative.

PR #264, opened 2018-05-10
Proposal for simpler composition code

danielmarbach:


adamralph:
@danielmarbach so I guess the difference here is that we add the original task to the pending tasks, instead of the conditional continuation task?

For my own education: I would have thought that the conditional continuation task would be completed in the case when it doesn't have to run, in which case it would satisfy the `Task.WhenAll()`. Is that not the case?

adamralph:
@danielmarbach I discussed this with @mauroservienti and we think we worked it out: the advantage of awaiting the original task is that we guarantee that the exception which walks up the stack is the exception thrown by the appender, and not an exception thrown by a logging framework. The continuation is marked to run synchronously, which means the `WhenAll()` still waits for it to run, but only has a reference to the append task, which guarantees that the correct exception walks up the stack. We return true from the continuation task because that in that context, we've done all we need to with the exception, i.e. logged it.

Do we get a gold ⭐️?

danielmarbach:
Yes. But ExecuteSynchronously is just a hint to the runtime. It is still
possible that under some circumstances the continuation is scheduled and
then the logging happens async. I figured the only "problem" this could
cause is if this rare case happens during shutdown and the logger is no
longer available (unlikely) then the continuation could throw but I guess
this is not really a big deal and almost unlikely to happen

Adam Ralph <notifications@github.com> schrieb am Do., 10. Mai 2018, 10:54:

mauroservienti:
Thanks @danielmarbach. I marked this as a SPIKE since it's far away from being done and we should discuss if we do really want to invest time in augmenting with logging all the 10 solutions (exercises + demos)

mauroservienti:
Ooops, no sorry is #263 that should be marked as a SPIKE.

I'm 👍 to merge this one into `add-logging`

adamralph:
@mauroservienti @danielmarbach any thoughts on my last comment?

danielmarbach:
Would definitely work your proposal. Which every we prefer is probably up
to someones gusto. I'm good with either or

Adam Ralph <notifications@github.com> schrieb am Do., 24. Mai 2018, 17:03:

adamralph:
I'll send a PR for that some time this week. It should be able to replace both this and #263

adamralph:
Raised #268 as an alternative.

PR #265, opened 2018-05-14
switch set-startup-projects.cmd for SwitchStartupProject

mauroservienti:
Visual Studio probably changed something in the `.suo` file format and now startup projects set using the `set-startup-projects` scripts are not preserved across Visual Studio restarts. Confirmed also by @DavidBoike that faced the same issue at NCD Minnesota last week.

I needed to find a workaround since as a trainer it happens that I re-open the same exercise multiple times. The easiest workaround is to use the [SwitchStartupProject for 2017](https://marketplace.visualstudio.com/items?itemName=vs-publisher-141975.SwitchStartupProjectforVS2017).

This si a WIP/SPIKE and is meant to be a workaround till we find a better solution, if any. I pushed and opened this just to raise awareness. And to have a backup 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,9 +81,9 @@ The exercises are contained in eight Visual Studio solutions under [exercises](e
 
 ## Running the exercise solutions
 
-Before opening any exercise solutions, set the startup projects by navigating to your copy of this repo and running `set-startup-projects.cmd`. Note that if you `git clean` your clone, you will have to run this command again.
+Before running an exercise solution, you need to set the startup projects.
 
-NOTE: do not run `set-startup-projects.cmd` while an exercise or demo solution is open. The script deletes and regenerates user settings for every solution. If it is run while a solution is open, the settings may be corrupted.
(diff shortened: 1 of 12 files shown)


adamralph:
@Particular/workshop-maintainers since I co-authored the commit, can I get another review please?

README.md
```diff
@@ -81,9 +81,9 @@ The exercises are contained in eight Visual Studio solutions under [exercises](e
 
 ## Running the exercise solutions
 
-Before opening any exercise solutions, set the startup projects by navigating to your copy of this repo and running `set-startup-projects.cmd`. Note that if you `git clean` your clone, you will have to run this command again.
+Before running an exercise solution, you need to set the startup projects.
 
-NOTE: do not run `set-startup-projects.cmd` while an exercise or demo solution is open. The script deletes and regenerates user settings for every solution. If it is run while a solution is open, the settings may be corrupted.
+The easiest way to do this is to use the [SwitchStartupProject](https://marketplace.visualstudio.com/items?itemName=vs-publisher-141975.SwitchStartupProjectforVS2017) Visual Studio extension. After installing the extension, open the exercise solution and choose the "Exercise" startup configuration.
```
danielmarbach: I don't think we should imply "easy".

How about switching it to something like

A way to do this is...

mauroservienti:
I cannot approve since I raised it, but I'm 👍

adamralph:
@mauroservienti if both you and I approve, that's 2 maintainers, so I think that's enough.

PR #273, opened 2018-06-28
switch to Bullseye and SimpleExec

adamralph:
- [Bullseye](https://github.com/adamralph/bullseye) replaces simple-targets-csx (it's a regular binary package instead of scripts, so can optionally be used in a [regular .NET project](https://github.com/xbehave/xbehave.net/blob/6d798877b4b891ce95ea8f2c5f9bee3d56a26bae/targets/Program.cs#L8)).
- [SimpleExec](https://github.com/adamralph/simple-exec/) wraps up command invocation, so removes the need to include local helpers


adamralph:
I've split the Bullseye switch into it's own PR: https://github.com/Particular/Workshop/pull/278

SimpleExec we can consider later.

PR #275, opened 2018-07-05
SqlLite and Entity Framework int64 issue

dvdstelt:
On the plane I tested the Composition Gateway demo and shipping had an issue where it could not retrieve data and crashed. I got an error saying `Expecting Int32 but I got an Int64 for Id column`, or the other way around.

I fixed it by adding a projection to exclude the `Id` column since we didn't use it anywhere, we only need the primary key for querying.

There might be a proper solution, but for now I just wanted to submit this so that if it fails for others, they are aware of this solution.

### DIFF
--- a/demos/asp-net-core/Divergent.Shipping.API.Host/Controllers/ShipmentsController.cs
+++ b/demos/asp-net-core/Divergent.Shipping.API.Host/Controllers/ShipmentsController.cs
@@ -20,6 +20,7 @@ namespace Divergent.Shipping.API.Host.Controllers
             {
                 var shipment = await db.Shipments
                     .Where(s => s.OrderNumber == orderNumber)
+                    .Select(s => new { s.OrderNumber, s.Courier, s.Status })
                     .SingleOrDefaultAsync();
 
                 return new


dvdstelt:
I don't, by now. I got a temporary laptop for SELA and in the plane I got this error. I couldn't Google it so I fixed the issue through this PR. By the time I got back I had to return the laptop. Only thing I know was that it was a 100% fresh install from a day before I traveled to SELA.

So other than this error I don't have anything.

adamralph:
@dvdstelt that's what I suspected. 😉 

I don't mind merging this, but we if we do, it follows that we should apply the same hack/fix/workaround to all the data projects, in both the demos and exercises. On the other hand, we could wait and see if this becomes a problem, or whether it was truly a one-off (i.e. very rare case). If we don't see it during the next few workshops, then we can close this PR. Thoughts?

dvdstelt:
I would say, better safe than sorry. Even if we can't reproduce the error, this will never hurt a single soul.

adamralph:
@dvdstelt I'm OK with that.

> if we do, it follows that we should apply the same hack/fix/workaround to all the data projects, in both the demos and exercises

On second thoughts, this is probably something specific to EF + SQLite, so we only have to do it for the demos. If you expand this PR to cover the `Sales` `OrdersController` then I believe we can merge.

dvdstelt:
@adamralph Not sure I'm following because we already have projections there.

adamralph:
@dvdstelt apologies, I missed that!

PR #286, opened 2018-08-16
switch build.cmd to dotnet run

adamralph:
Since the [switch to Bullseye](https://github.com/Particular/Workshop/pull/278), we don't have to use C# scripting any more. We can just use `dotnet run`.


adamralph:
@Particular/workshop-maintainers anyone have time to review this?

FTR, this is how most projects are using Bullseye now:

- https://github.com/blairconrad/SelfInitializingFakes/blob/master/tools/targets/Program.cs
- https://github.com/FakeItEasy/FakeItEasy/blob/develop/tools/FakeItEasy.Build/Program.cs
- https://github.com/SQLStreamStore/SQLStreamStore/blob/master/build/Program.cs
- https://github.com/xbehave/xbehave.net/blob/master/targets/Program.cs
- https://github.com/adamralph/liteguard/blob/master/targets/Program.cs

It's less overhead, tidier, and has more benefits, than using scripting. It would be nice to get this in before Explore DDD and NDC Sydney.

adamralph:
FYI I amended the commit to use Bullseye 1.2.0 instead of 1.1.0-rc.2.

PR #290, opened 2018-08-23
[WIP] Fix wrong path in the setup database ps script

matteobaglini:


mauroservienti:
I always hit this. I prefer @matteobaglini's approach, that requires shell to be in the scripts directory.

adamralph:
Note that if the set up instructions are followed, this works as-is, see https://github.com/Particular/Workshop#set-up-the-databases

> Open an **elevated** command prompt, navigate to your copy of this repo, and run:
> 
> ```Batchfile
> powershell -NoProfile -ExecutionPolicy unrestricted -File exercises\scripts\Setup-LocalDBInstance.ps1
> ```

You'll only encounter problems if you deviate from these set up instructions. That is also the reason that we cannot merge this PR as-is, since it would be in conflict with the set up instructions.

That said, I'd like to support either scenario. I'll look into it and see if that's possible.

adamralph:
#290 allows either scenario.

adamralph:
Thanks again for this PR @matteobaglini. Although it wasn't merged, it prompted us to solve the problem for both scenarios over in #291.

matteobaglini:
Thank you @adamralph  for the feedback and the fix.

Issue #295, opened 2018-09-03
Minimize HTTP API console windows


timbussmann:
Outcome of the discussion in https://github.com/Particular/Workshop/issues/280
Consider minimizing the console windows of the HTTP API processes to minimize the visual overhead to users as they most likely won't need to interact with those console windows.

mauroservienti:
One other option could be to affect the way apps are grouped in the taskbar by calling `SetCurrentProcessExplicitAppUserModelID`.

Ref. https://docs.microsoft.com/en-us/windows/desktop/api/shobjidl_core/nf-shobjidl_core-setcurrentprocessexplicitappusermodelid

PR #297, opened 2018-09-03
[wip] [spike] Minimize HTTP API console windows

adamralph:
Connects to #295.

This will do it.  Note that it needs `net471` or higher, or a switch to any version of `netcoreapp`: https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.runtimeinformation#applies-to

If we like it, we need to think about how to get this code into all the API projects. I'm reluctant to duplicate it.

### DIFF
--- a/exercises/04-integration/after/Divergent.Customers.API/ConsoleEx.cs
+++ b/exercises/04-integration/after/Divergent.Customers.API/ConsoleEx.cs
@@ -1,0 +1,15 @@
+﻿using System.Runtime.InteropServices;
+
+namespace Divergent.Customers.API
+{
+    static class ConsoleEx
+    {
+        public static void TryMinimize()
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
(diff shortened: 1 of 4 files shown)


exercises/04-integration/after/Divergent.Customers.API/ConsoleEx.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System.Runtime.InteropServices;
+
+namespace Divergent.Customers.API
+{
+    static class ConsoleEx
+    {
+        public static void TryMinimize()
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
timbussmann: is that check really necessary when targeting the .NET Framework?

adamralph: @timbussmann it's an edge case, but you could run it on Mono on *nix. But in any case, we want to move to .NET Core ASAP anyway, so I'd rather include this check now.

adamralph:
I'll raise a proper PR to get this in if/when the opportunity arises.

PR #313, opened 2018-11-29
Applying the Composite also to a ViewModel generated by the MVC controller

BWalti:
Beeing able to use strongly typed "frontend models" as well as to set some "default values", would probably be an interesting possibility.

### DIFF
--- a/demos/asp-net-core/Divergent.Website/Controllers/OrdersController.cs
+++ b/demos/asp-net-core/Divergent.Website/Controllers/OrdersController.cs
@@ -11,7 +11,21 @@ namespace Divergent.Website.Controllers
 
         public IActionResult Details(int id)
         {
-            return View();
+            return View(new OrderDetail
+                            {
+                                ShippingStatus = "Delivered",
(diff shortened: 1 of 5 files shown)


BWalti:
I just realized that this wouldn't work for collection properties. One way to solve this would be to serialize the dynamic view model into json and deserialize it again into the "input viewmodel type". But this would be insane, as it would directly be serialized again.. Thats an overhead which probably can't be justified.
Have it strongly typed from the beginning could probably be achieved too, but still..

mauroservienti:
Thanks @BWalti for raising this. As you noticed introducing strongly typing on top of the current infrastructure is not that easy. And to be honest is not the goal here. The goal here is to provide a sample that does the job with a low entry barrier. I have a spike that is based on strong typing only, the problem is that it makes the infrastructure code explode and it requires a lot of interfaces on the ViewModel classes just to keep the logical coupling low.

I personally don't think there is much value in strong typing when it comes to ViewModel Composition. Obviously I may be wrong 😄

PR #320, opened 2019-03-28
Ex2 tweaks

tmasternak:


exercises/03-sagas/README.md
```diff
@@ -127,15 +127,31 @@ class ShippingSagaData : ContainSagaData
 }
 ```
 
+NOTE: One could also use a property of type `int[]` for holding product identifiers as it's supported by the SqlPersister used in this exercise.
```
adamralph: Maybe we should go ahead and change the type instead? Otherwise we should really justify our use of `ICollection<int>` over `int[]`.

exercises/03-sagas/README.md
```diff
@@ -127,15 +127,31 @@ class ShippingSagaData : ContainSagaData
...
 
 Now we'll add this class as saga state to our saga. We'll end up with a class like this:
 
 ```c#
-class ShippingSaga : Saga<ShippingSagaData>,
-    IAmStartedByMessages<OrderSubmittedEvent>,
-    IAmStartedByMessages<PaymentSucceededEvent>
+public class ShippingSaga : Saga<ShippingSagaData>,
+        IAmStartedByMessages<OrderSubmittedEvent>,
+        IAmStartedByMessages<PaymentSucceededEvent>
```
adamralph: These extra tabs seem unecessary. I would keep them as-is.

exercises/03-sagas/README.md
```diff
@@ -127,15 +127,31 @@ class ShippingSagaData : ContainSagaData
...
+
+    public Task Handle(OrderSubmittedEvent message, IMessageHandlerContext context)
+    {
+        throw new System.NotImplementedException();
+    }
+
+    public Task Handle(PaymentSucceededEvent message, IMessageHandlerContext context)
+    {
+        throw new System.NotImplementedException();
+    }
```
adamralph: I think this has been suggested before. I'm not sure that the code needs to be compilable between each step but it's true that the current "Step 4" suggests it should be. I'm not sure it's worth adding this code only to remove it again, but I guess `Ctrl+.` can do it for us anyway.

Should we go ahead with this, or should we remove step 4 (or move it later)?

exercises/03-sagas/README.md
tmasternak: I'm 👍 for using `int[]`

exercises/03-sagas/README.md
tmasternak: The thing that was misleading to me was that we show:

```
public class ShippingSaga : Saga<ShippingSagaData>,
    IAmStartedByMessages<OrderSubmittedEvent>,	        
    IAmStartedByMessages<OrderSubmittedEvent>,
    IAmStartedByMessages<PaymentSucceededEvent>
{
}
```

this includes both `{` and `}` which suggest that we look at the complete class definition.

exercises/03-sagas/README.md
adamralph: IMO you can go ahead and make that change instead.

exercises/03-sagas/README.md
adamralph: Reading back through the steps, I think it may be worth considering re-writing most or even all of them. Some of them don't make much sense to me. E.g.

> The class should now look like this:
> 
> ```C#
> class ShippingSaga : Saga<object>,
>     IAmStartedByMessages<OrderSubmittedEvent>,
>     IAmStartedByMessages<PaymentSucceededEvent>
> {
> }
> ```
> 
> Override saga base class `ConfigureHowToFindSaga abstract` method and implement both interfaces, leaving methods bodies as is. Before we can implement these, we need to add some saga state.

What does "implement both interfaces, leaving methods bodies as is" mean?

dvdstelt: > implement both interfaces

Means you need to implement them, make sure the code compiles?

> leaving methods bodies as is

Could also be understood/rewritten as "leave the method body empty"

I'm not sure why this is suddenly an issue, because afaik this has been there since the beginning. I've never heard anyone ask questions or complain about it. It seems like we're fixing something that isn't broken, which feels gold plating.

adamralph: 👍 to leaving this as-is for now.

